Issue #5, opened 2017-04-20
Tests fail on nightly


imp:
Looks like `cargo` that comes with nightly renders metadata a bit differently and causes test failure

oli-obk:
I think we need to make some fields optional to be able to support multiple cargo versions at once

oli-obk:
Works now

PR #7, opened 2017-04-20
Fix tests on nightly

imp:
Addresses #5


imp:
Looks like there is a breaking change between beta and nightly `cargo metadata` output. If one passes the test then the other fails. Bummer.

oli-obk:
Hmm unfortunate. Can you pinpoint the exact discrepancy?

imp:
Sure. For tests the value of `crate_types` is "bin" on nightly and "test" on beta.

PR #9, opened 2017-06-11
 Include dependencies graph in the `Metadata`

mcarton:
I plan to use this for https://github.com/Manishearth/rust-clippy/issues/1797.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,8 +23,25 @@ use std::io;
 pub struct Metadata {
     /// A list of all crates referenced by this crate (and the crate itself)
     pub packages: Vec<Package>,
-    resolve: Option<()>,
+    /// Dependencies graph
+    pub resolve: Option<Resolve>,
     version: usize,
+}
+
+#[derive(Clone, Deserialize, Debug)]
+/// A dependency graph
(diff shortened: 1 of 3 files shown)


oli-obk:
Why a major bump? This only makes previously private fields public.

src/lib.rs
```diff
@@ -23,25 +23,43 @@ use std::io;
 pub struct Metadata {
     /// A list of all crates referenced by this crate (and the crate itself)
     pub packages: Vec<Package>,
-    resolve: Option<()>,
+    /// Dependencies graph
+    pub resolve: Option<Resolve>,
```
mcarton: I bumped the version because this type changes.

oli-obk: Well. The field was private, so it's only a small breaking change since the parsed json is tested more strictly. I think it should be a patch version bump, since no real code should break (it should be valid metadata json to begin with)

mcarton: Oh right, don't know what I was thinking there.

oli-obk:
Thx, will publish on Thursday. Earlier if I have a PC and internet simultaneously

Issue #3, opened 2017-04-20
`workspace_members` field of `cargo metadata` is not supported


imp:
It looks like `workspace_members` field is not exposed despite being reported by `cargo metadata`

oli-obk:
This works now

PR #10, opened 2017-07-13
add `workspace_members` field to solve #3

flier:
add `workspace_members` field to solve #3

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,8 @@ use std::io;
 pub struct Metadata {
     /// A list of all crates referenced by this crate (and the crate itself)
     pub packages: Vec<Package>,
+    /// A list of all workspace members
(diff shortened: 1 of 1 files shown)


oli-obk:
Is this a new field? As in was there a cargo version that didn't report it? If so, please add `#[serde(default)]` so the field also works on older compiler versions

flier:
yes, it contains the workspace members, like

```json
{
  "workspace_members": [
    "mockit 0.1.0 (path+file:///Users/flier/github/rust-mockit/mockit)",
    "mockit-derive 0.1.0 (path+file:///Users/flier/github/rust-mockit/mockit-derive)"
  ],
  "resolve": null,
  "version": 1
}
```

oli-obk:
Thanks!

Issue #11, opened 2017-08-21
impl Error for cargo_metadata::Error?


bjgill:
It would make using `cargo_metadata::Error` substantially easier.

Would it make sense, perhaps, to switch over to [`error-chain`](https://crates.io/crates/error-chain)?

oli-obk:
Definitely! I don't have much experience with `error-chain`, but I'll happily accept PRs ;)

bjgill:
Fixed in #15

PR #15, opened 2017-08-22
Use error_chain; add/fix manifest_path support; run rustfmt; expose extra `Dependency` fields

bjgill:
This PR does a few things:
* Use `error_chain` for errors - #11 
* Add/fix support for providing manifest path
* Run the latest rustfmt-nightly
* Make public the `req` and `kind` fields on `Dependency`

(happy to split up if that would make reviewing easier).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
-name = "cargo_metadata"
-version = "0.2.3"
 authors = ["Oliver Schneider <git-spam-no-reply9815368754983@oli-obk.de>"]
-repository = "https://github.com/oli-obk/cargo_metadata"
 description = "structured access to the output of `cargo metadata`"
 license = "MIT"
-
+name = "cargo_metadata"
+repository = "https://github.com/oli-obk/cargo_metadata"
+version = "0.2.3"
 [dependencies]
+error-chain = "0.10.0"
 serde = "1.0.2"
-serde_json = "1.0.1"
 serde_derive = "1.0.2"
+serde_json = "1.0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,8 +5,9 @@ license = "MIT"
 name = "cargo_metadata"
 repository = "https://github.com/oli-obk/cargo_metadata"
 version = "0.2.3"
+
 [dependencies]
 error-chain = "0.10.0"
-serde = "1.0.2"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
-name = "cargo_metadata"
-version = "0.2.3"
 authors = ["Oliver Schneider <git-spam-no-reply9815368754983@oli-obk.de>"]
-repository = "https://github.com/oli-obk/cargo_metadata"
 description = "structured access to the output of `cargo metadata`"
 license = "MIT"
-
+name = "cargo_metadata"
```
oli-obk: What happened here? These changes look rather unrelated. I can understand ordering the dependencies by name, but the package entries should stay in the idiomatic order.

src/lib.rs
```diff
@@ -151,8 +132,8 @@ pub fn metadata_deps(manifest_path_arg: Option<&str>, deps: bool) -> Result<Meta
     if let Some(mani) = manifest_path_arg {
         cmd.arg(mani);
     }
-    let output = cmd.output()?;
-    let stdout = from_utf8(&output.stdout)?;
-    let meta: Metadata = serde_json::from_str(stdout)?;
+    let output = cmd.output().chain_err(|| "Failed to execute `cargo metadata`")?;
```
oli-obk: hmm... I actually like that we had the real error in here. Can you reintroduce the variants via error_chain?

Cargo.toml
```diff
@@ -5,8 +5,9 @@ license = "MIT"
 name = "cargo_metadata"
 repository = "https://github.com/oli-obk/cargo_metadata"
 version = "0.2.3"
+
 [dependencies]
 error-chain = "0.10.0"
-serde = "1.0.2"
-serde_derive = "1.0.2"
-serde_json = "1.0.1"
+serde = "1.0.11"
```
oli-obk: The version bump is unnecessary, right?

Cargo.toml
bjgill: Ah - using a tool to add `error_chain`. It has a tendency to mangle manifests.

bjgill:
@oli-obk - have made the changes you suggest. I've also gone on to exposing the pair of fields on `Dependency` that I was wanting to use from another crate.

oli-obk:
Thanks! This looks great now

Issue #16, opened 2017-08-25
Breaking change to handling of `manifest_path_arg`?


bjgill:
As of #15, `metadata()` still takes an `Option<&str>`, but expects it to be of the form e.g. `Some("Cargo.toml")`. See https://github.com/oli-obk/cargo_metadata/blob/master/tests/selftest.rs#L21

Looking at the example, though, I think #15 may have introduced a breaking change. Unless I'm misreading the example (and code), it was previously expecting something of the form e.g. `Some("--manifest_path=Cargo.toml")`.

```rust
let manifest_path_arg = std::env::args().skip(2).find(|val| val.starts_with("--manifest-path="));
let metadata = cargo_metadata::metadata(manifest_path_arg.as_ref().map(AsRef::as_ref)).unwrap();
```

More irritating is that this is the sort of breaking change that nobody will notice until they try and pass in a manifest path - there have been no changes to the types.

I don't know what the right answer is. Maybe it's worth reverting this change? Maybe we could switch to `Option<Path>` so that people notice the change when their code fails to compile?

Also, I'm not convinced the old code (or example) supports providing the manifest path as `--manifest-path PATH` (i.e. with the '`=`' optional, which is what `cargo` does using `docopt`).

oli-obk:
Yea we'll do a major version bump. There are a few breaking changes anyway. We should definitely change it to Path though, yes

bjgill:
OK - I'll look at sorting that out soon.

I'd also like to add examples of using this with `docopt` and `clap`

PR #17, opened 2017-08-26
Metdata path is now a Path; add examples of use

bjgill:
Add a readme and usage examples. Also update the manifest path input to be an `Option<&Path>`

Should solve #16


oli-obk:
Awesome! Thanks. Those are some great docs

Issue #12, opened 2017-08-21
Expose Dependency kind and version requirement?


bjgill:
I'm hoping to use `cargo_metadata` for [`cargo-edit`](https://github.com/killercup/cargo-edit) to get a full list of all the direct dependencies of crates in a workspace.

For this to be useful, however, it would be nice if `Dependency` were to expose the kind and version.

oli-obk:
Just open a PR to make fields public that you need, I simply kept all fields private that I didn't need. If a struct's fields all become public, add a private field that is defaulted, so we can keep extending the struct in the future without major version bumps.

bjgill:
It would be nice to be able to have the dependency version as a `VersionReq` (rather than a `String`). However, I think https://github.com/steveklabnik/semver/issues/119 currently blocks this.

oli-obk:
You can always use `deserialize_with` to work around that.

bjgill:
I'm trying to fix it at source with https://github.com/steveklabnik/semver/pull/122

bjgill:
OK - semver can now deserialize `VersionReq`, so this will be trivial to fix once a new version of semver is released.

oli-obk:
Once we get this, I'll publish a new major version.

oli-obk:
:tada: Do you have any other breaking changes you wanted to get into the next major release?

bjgill:
I don't think so. #19 is the only issue outstanding, and it doesn't seem particularly important.

bjgill:
@oli-obk - when were you planning to publish the new version?

oli-obk:
ups. on it. sorry!

oli-obk:
Published 0.3

bjgill:
Thank you :tada:

PR #18, opened 2017-08-28
Use strongly-typed version requirements

bjgill:
`semver` now allows deserialization of `VersionReq`, so we can use that
in `Dependency` (instead of a plain `String`).

Finishes solving #12

### DIFF
--- a/tests/selftest.rs
+++ b/tests/selftest.rs
@@ -37,4 +37,13 @@ fn metadata_deps() {
     assert_eq!(this.targets[1].name, "selftest");
     assert_eq!(this.targets[1].kind[0], "test");
     assert_eq!(this.targets[1].crate_types[0], "bin");
+
+    let dependencies = &this.dependencies;
+
+    let serde = dependencies
+        .iter()
+        .find(|dep| dep.name == "serde")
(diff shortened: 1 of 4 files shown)


tests/selftest.rs
```diff
@@ -37,4 +37,13 @@ fn metadata_deps() {
...
     assert_eq!(this.targets[1].crate_types[0], "bin");
+
+    let dependencies = &this.dependencies;
+
+    let serde = dependencies
+        .iter()
+        .find(|dep| dep.name == "serde")
+        .expect("Did not find serde dependency");
+
+    assert_eq!(serde.kind, cargo_metadata::DependencyKind::Normal);
```
oli-obk: I think you should also add a test for the version. Something like checking if the serde dependency has a version `> 1.0.0`.

bjgill: We currently demand that `serde` should be `"1.0.2"` - i.e. `^1.0.2 := >=1.0.2 <2.0.0`

I've added checks for the upper and lower bounds, and that a legal value is accepted.

Of course, this will break if we ever update to semver 2.0.0, but I suspect a semver 2.0.0 is unlikely (and easy to fix if so).

PR #21, opened 2017-09-15
Export ErrorKind

bjgill:
Needed to allow error chaining to work properly.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,13 @@
+//! Create the `Error`, `ErrorKind`, `ResultExt`, and `Result` types
+#![allow(missing_docs)] // see https://github.com/rust-lang-nursery/error-chain/issues/63
+
+error_chain!{
+    foreign_links {
+        // Error during execution of `cargo metadata`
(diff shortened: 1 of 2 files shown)


bjgill:
Sorry - a rather silly omission on my part when switching to error chain.

oli-obk:
Travis is unhappy with your changes without docs

bjgill:
Unfortunately, due to https://github.com/rust-lang-nursery/error-chain/issues/63, I don't think I can add docs to the foreign links.

I think there are two possible solutions:

1. move the `error_chain!{..}` to an `errors` module, and mark that as `![allow(missing_docs)]`
2. remove the `foreign_links` section, and just use `.chain_err(|| "...")` instead

Thoughts? I think I prefer the second solution, as it seems like a more natural use of error-chain.

oli-obk:
Sounds to me like the second solution is a breaking change.

bjgill:
Ok - the first is easy enough to implement

oli-obk:
I'll publish as soon as I get to a PC... which might not be before Friday next week though

PR #22, opened 2017-10-31
Use doc attributes for foreign links

bjgill:
Uglier than before, but at least rustdoc picks up the comments now so we can remove the `allow`.


oli-obk:
Ugly hack, but whatever works, right? :D

Thanks

PR #24, opened 2017-11-03
Expose `features` on Package.

### LINKED ISSUE #23: Expose `features` for `Package`s
This field is currently present, but non-public. Could it be made public?

Mark-Simulacrum:
Adds another field which is skipped by Serde to ensure that users cannot
match the struct exhaustively.

Fixes #23.


oli-obk:
Thanks! Published as 0.3.2

PR #25, opened 2017-11-27
Make source field public

topecongiro:
I would like to make the source field public to see if the dependency is local or not.


oli-obk:
Thanks.

Published as 0.3.3

PR #26, opened 2017-12-20
Add WorkspaceMember

topecongiro:
The implementation of WorkspaceMember is almost identical to PackageId in cargo.
The difference is that its deserialization is simplified and it does not use Arc internally.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,7 @@ use std::str::from_utf8;
 
 pub use errors::{Error, ErrorKind, Result};
 pub use dependency::{Dependency, DependencyKind};
+use serde::de;
 
 mod errors;
 mod dependency;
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "MIT"
 readme = "README.md"
 
 [dependencies]
+cargo = "0.23"
```
oli-obk: I'd rather be independent from the cargo crate. Can we just use `String` instead of `SourceId` for now?

topecongiro:
Thank you for the review, I replaces `SourceId` with `String`.

oli-obk:
Ok, so this'll be a breaking change. I'll check whether there are other things I wanted to break and then release a new version

oli-obk:
Published 0.4.0

The only code that breaks is code exhaustively matching on the structs.

oli-obk:
Maybe we should add a `Display` impl that reproduces the original `String`, so anyone who used the `workspace_member` before can just keep using it by calling `to_string()`

PR #27, opened 2018-01-06
Implement `Serialize` on everything

koute:


oli-obk:
Can you supply a motivation and mention it in the readme, too?

koute:
The motivation is mainly to support generating `--message-format=json` output in `cargo` subcommands just as `cargo` does. Some of the types in what `cargo --message-format=json` emits are exactly the same as the ones from `cargo metadata`.

How exactly would you like this mentioned in the README? Would this be fine?

> Also supports serialization to aid in implementing `--message-format=json`-like output generation in `cargo-*` subcommands, since some of the types in what `cargo --message-format=json` emits are exactly the same as the ones from `cargo metadata`.

oli-obk:
Jup, that's perfect.

I just didn't know about that use case.

koute:
README updated! Is that fine?

oli-obk:
Great! thanks. I'll push a new minor version asap

oli-obk:
Published as 0.4.1

PR #31, opened 2018-03-11
Add target_directory to Metadata

AndrewGaspar:


oli-obk:
Thx, I published a new version that includes this

PR #36, opened 2018-05-30
handle cargo features flags

gibix:
Not sure about the API, if sounds good I can write an example before merging.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,6 +72,21 @@
 //! let _metadata =
 //!     cargo_metadata::metadata(matches.value_of("manifest-path").map(Path::new)).unwrap();
 //! ```
+//!
+//! Pass features flags
+//!
+//! ```rust
+//! # // This should be kept in sync with the equivalent example in the readme.
+//! # extern crate cargo_metadata;
+//! # use std::path::Path;
+//!
+//! let manifest_path = Path::new("./Cargo.toml");
+//! let features = cargo_metadata::CargoOpt::AllFeatures;
+
+//! let _metadata =
+//! cargo_metadata::metadata_run(Some(manifest_path), false, Some(features)).unwrap();
+//!
+//! ```
 
 #[macro_use]
 extern crate error_chain;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -247,6 +270,13 @@ pub fn metadata_deps(manifest_path: Option<&Path>, deps: bool) -> Result<Metadat
         cmd.arg("--no-deps");
     }
 
+    match feat {
+        CargoOpt::AllFeatures => cmd.arg("--features-all"),
+        CargoOpt::NoDefaultFeatures => cmd.arg("--no-default-features"),
+        CargoOpt::SomeFeatures(ftrs) => cmd.arg(format!("--fatures {:?}", ftrs)),
```
oli-obk: typo: `fatures`

src/lib.rs
```diff
@@ -247,6 +270,13 @@ pub fn metadata_deps(manifest_path: Option<&Path>, deps: bool) -> Result<Metadat
...
+        CargoOpt::DefaultFeatures => cmd.arg(""),
```
oli-obk: This arm should be a nop `{}` arm instead of adding an empty argument

gibix: is not possible to give a nop arg because cmd.arg returns a Command. Suggestions?

oli-obk: Just make the last arm `cmd` instead of `cmd.arg("")`, that should do it.

gibix:
changed to Option<CargoOpt>`

src/lib.rs
```diff
@@ -247,6 +268,14 @@ pub fn metadata_deps(manifest_path: Option<&Path>, deps: bool) -> Result<Metadat
         cmd.arg("--no-deps");
     }
 
+    if features.is_some() {
```
oli-obk: `if let Some(features) = features` saves you the unwrap

gibix: done, thanks for the feedback, is a rust-beginner contribution ;)

oli-obk:
>  I can write an example before merging.

Code looks good to me! Can you write this example so we have some form of testing for this feature?

gibix:
ping

oli-obk:
0.5.6 has been published with this feature

gibix:
thanks

PR #37, opened 2018-05-31
Add support for package.metadata table

phil-opp:
Since https://github.com/rust-lang/cargo/pull/5360 `cargo metadata` outputs the contents of the [optional `package.metadata` table](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table-optional). This PR adds support for the table to this crate. Since the table can contain arbitrary content, the return type is a dynamic `serde_json::Map`.

Unresolved questions:

- Is it ok to expose types of the serde_json dependency in the public API? Or do we need to make the `serde_json` dependency public?
- The test doesn't work on stable yet. Is there a way to conditionally execute the test only on nightly and beta?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,6 +155,8 @@ pub struct Package {
     pub features: HashMap<String, Vec<String>>,
     /// Path containing the `Cargo.toml`
     pub manifest_path: String,
+    /// Contents of the free form package.metadata section
+    pub metadata: Option<serde_json::Map<String, serde_json::Value>>,
     #[doc(hidden)]
     #[serde(skip)]
     __do_not_match_exhaustively: (),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -156,6 +156,7 @@ pub struct Package {
     /// Path containing the `Cargo.toml`
     pub manifest_path: String,
     /// Contents of the free form package.metadata section
+    #[serde(default)]
     pub metadata: serde_json::Value,
     #[doc(hidden)]
     #[serde(skip)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -156,6 +156,30 @@ pub struct Package {
     /// Path containing the `Cargo.toml`
     pub manifest_path: String,
     /// Contents of the free form package.metadata section
+    ///
+    /// This contents can be serialized to a struct using serde:
+    ///
+    /// ```rust
+    /// #[macro_use]
+    /// extern crate serde_json;
+    /// #[macro_use]
+    /// extern crate serde_derive;
+    ///
+    /// #[derive(Debug, Deserialize)]
+    /// struct SomePackageMetadata {
+    ///     some_value: i32,
+    /// }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -155,6 +155,8 @@ pub struct Package {
     pub features: HashMap<String, Vec<String>>,
     /// Path containing the `Cargo.toml`
     pub manifest_path: String,
+    /// Contents of the free form package.metadata section
+    pub metadata: Option<serde_json::Map<String, serde_json::Value>>,
```
oli-obk: Does this need to be an `Option`? Can't it simply be an empty table? I don't want to distinguish between no table and an empty table. That feels like noise

phil-opp:
> There should be env vars in travis.

Ok. Should I add an `if` condition to the test that checks the environment variable? Or should I create a new nightly/beta only test and put the condition in the `travis.yml`?

> Maybe we could make the field generic. If it's generic the user can choose their own metadata type instead of having to get their data from a serde map.

Sounds like a good idea, but I'm not quite sure if this would work with multiple packages. The `metadata` struct stores a `Vec<Package>`, so if we make `Package` generic all packages must have the same metadata package.

src/lib.rs
phil-opp: An empty table doesn't parse if the key is not present. But I pushed a new version that makes the field a `serde_json::Value` (null if not present).

oli-obk: Have you tried `#[serde(default)]` on the `metadata` field?

oli-obk:
> Ok. Should I add an if condition to the test that checks the environment variable? Or should I create a new nightly/beta only test and put the condition in the travis.yml?

just put it in the test, maybe with a nice error message if the condition is not set, so users running locally know what's going on.

> Sounds like a good idea, but I'm not quite sure if this would work with multiple packages. The metadata struct stores a Vec<Package>, so if we make Package generic all packages must have the same metadata package.

oh god. yea let's not do anything here. Just store the json and maybe provide an example in the readme showing how to deserialize that into a custom structure.

src/lib.rs
phil-opp: It seems like the metadata field is always present and `null` if not set. `serde(default)` fails in this case. So I think `serde_json::Value` is the correct type here.

phil-opp: But `serde(default)` helps that no “missing field `metadata`” error occurs on stable.

phil-opp:
I added a “NOTE: This test currently only works on the beta and nightly channel.” error message and a short example how to transform a `serde_json::Value` into a custom struct.

phil-opp:
Thanks! Could you release a new version to crates.io?

oli-obk:
I do have a deskop open for it and even ran `cargo test` then forgot about it -.- Thanks for the reminder

oli-obk:
published as v0.5.5

phil-opp:
Thanks!

PR #38, opened 2018-07-01
Re-export backtrace feature of error-chain

alecmocatta:
On Linux and other targets, the `backtrace` crate relies upon the `cc` crate to compile `libbacktrace` – a C library. This can complicate the build.

This pull request re-exposes the `backtrace` feature of `error-chain` such that it can optionally be disabled.


oli-obk:
published a new version. Would you be so kind and add something to travis.yml so we test both the version with and without the feature?

PR #40, opened 2018-07-02
add authors key to `Package`

euclio:


oli-obk:
Thanks!

Published as 0.5.8

PR #41, opened 2018-07-10
Expose remaining fields in Dependency

zachreizner:
The private fields in Dependency are required to perform precise dependency resolution and there is no need to keep this fields private. This change makes all the fields public and gives them a short doc, except of course __do_not_match_exhaustively.

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -46,9 +46,12 @@ pub struct Dependency {
     pub kind: DependencyKind,
     /// Whether this dependency is required or optional
     pub optional: bool,
-    uses_default_features: bool,
-    features: Vec<String>,
-    target: Option<String>,
+    /// Whether the default features in this dependency are used.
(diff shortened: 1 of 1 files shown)


zachreizner:
The structure of the target string seems complex and ill-defined. The only reference I can even find to its structure is in this [pull request](https://github.com/rust-lang/cargo/pull/2328). I cranked out a list of unique targets I could find on crates.io [here](https://gist.github.com/zachreizner/b5cef01143774544dc6f412f8ae44c70). Maybe it would make sense to parse that, but it may be brittle as it seems partially defined by whatever `rustc --print cfg` happens to dump. I also don't necessarily need the target field for what I'm working (a fancy version of cargo-local-registry), so I would be willing to punt this problem by keeping that particular field private.

oli-obk:
I'll publish tomorrow. I'm on mobile right now

PR #43, opened 2018-07-10
add keys for license and license_file to `Package`

euclio:


src/lib.rs
```diff
@@ -159,6 +159,13 @@ pub struct Package {
     pub name: String,
     /// Version given in the `Cargo.toml`
     pub version: String,
+    /// The [SPDX license expression][spdx-expressions] for this package.
+    ///
+    /// [spdx-expressions]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60
+    pub license: Option<String>,
+    /// If the package is using a nonstandard license, this key may be specified instead of
+    /// `license`, and must point to a file relative to the manifest.
+    pub license_file: Option<String>,
```
oli-obk: This should probably be a `PathBuf` then

src/lib.rs
```diff
@@ -159,6 +159,13 @@ pub struct Package {
     pub name: String,
     /// Version given in the `Cargo.toml`
     pub version: String,
+    /// The [SPDX license expression][spdx-expressions] for this package.
+    ///
+    /// [spdx-expressions]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60
```
oli-obk: I have no idea about this topic. But doesn't that mean that we might in the future expose a more structured way to access the license (so not just a `String`)?

If so, please create a `License` type that contains a private `String` field and exposes that via a `Display` impl.

src/lib.rs
euclio: I didn't make it a PathBuf because you can technically store any string in this field, not necessarily a path. I could make it a PathBuf, but I suppose there's a risk of it not deserializing correctly.

oli-obk: We already have a bunch of things that can go awry. Let's assume ppl use this field somewhat sensibly

oli-obk:
ping @euclio

euclio:
Looks like `license` got added in the interim so I just added `license_file`.

mckaymatt:
Any chance this will be released soon? I'd like to use the `license_file`.

oli-obk:
woop sorry, published 0.6.4

PR #47, opened 2018-08-08
Accept `AsRef<Path>` instead of just `Path`

LegNeato:
This makes it more ergonomic for library users.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -264,7 +264,7 @@ pub enum CargoOpt {
 /// # Parameters
 ///
 /// - `manifest_path`: Path to the manifest.
-pub fn metadata(manifest_path: Option<&Path>) -> Result<Metadata> {
+pub fn metadata<P: AsRef<Path>>(manifest_path: Option<P>) -> Result<Metadata> {
     metadata_run(manifest_path, false, None)
 }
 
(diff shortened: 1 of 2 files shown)


tests/selftest.rs
```diff
@@ -21,7 +21,7 @@ struct TestPackageMetadata {
 
 #[test]
 fn metadata() {
-    let metadata = cargo_metadata::metadata(None).unwrap();
+    let metadata = cargo_metadata::metadata(None::<&Path>).unwrap();
```
oli-obk: hm... this is a breaking change, and the `None` case seems like a very common case

I'll hold off until the next breaking change

src/lib.rs
```diff
@@ -264,7 +264,7 @@ pub enum CargoOpt {
 /// # Parameters
 ///
 /// - `manifest_path`: Path to the manifest.
-pub fn metadata(manifest_path: Option<&Path>) -> Result<Metadata> {
+pub fn metadata<P: AsRef<Path>>(manifest_path: Option<P>) -> Result<Metadata> {
```
oli-obk: can you use `Option<impl AsRef<Path>>` instead of explicit generic args?

LegNeato: That would set the minimum rust version to 1.26 ...is that ok?

oli-obk: Since this is only happening with the next breaking change, ppl will need to manually bump their version anyway, so seems fine.

tests/selftest.rs
oli-obk: Maybe we should just have two different functions instead of an `Option` argument? What do you think about `metadata` acting like the `None` variant and having no args and a `metadata_for_manifest` function that just takes an `AsRef<Path>`?

LegNeato: I'm fine with that if that's what you want!

PR #49, opened 2018-08-17
Add edition field.

### LINKED ISSUE #48: Support `edition`
`cargo metadata` now emits edition information (see rust-lang/cargo#5834).  

I'm not sure what your policy is for nightly features, but I'd be happy to add this if you want.

ehuss:
Closes #48.


ehuss:
I'm not sure if you'd like tests for this.  Most of the other fields didn't have any tests, and to properly test this would require a separate manifest file, and limiting it to nightly.  I did run some tests locally and it seems to work ok.

PR #50, opened 2018-08-25
Bump serde dependency version

mikerite:
This crate uses the `serde(transparent)` container attribute but this
attribute was only added in serde v1.0.59.


oli-obk:
good catch, thx

PR #52, opened 2018-09-02
Add ability to set cwd

matklad:
This might be useful to get "find nearest Cargo.toml in an ancestor directory" behavior for free.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,7 +84,7 @@
 //! let features = cargo_metadata::CargoOpt::AllFeatures;
 
 //! let _metadata =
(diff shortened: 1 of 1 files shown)


oli-obk:
I'll include this in the next major version bump

oli-obk:
This PR needs a rebase

matklad:
I think I should use `Option<impl AsRef<Path>>` as well, but it makes the code ugly due to https://github.com/oli-obk/cargo_metadata/pull/47#discussion_r208477873.

I'll try to draft a builder-based interface instead of the current bunch of funcitons.

PR #53, opened 2018-09-10
Add message-format=json support

### LINKED ISSUE #45: Add structs output by `--message-format=json`?
I'm using the following to parse `cargo build --message-format=json`. I see other crates have a similar redefinition of this, for example [assert_cmd](https://github.com/assert-rs/assert_cmd/blob/4cf3ef050fd0ddfbdfed788652a24590e69f6a5c/src/cargo.rs#L93-L125).

They are [apparently](https://users.rust-lang.org/t/how-stable-is-cargos-message-format-json/15662) stable. I'm wondering if it's worth making cargo_metadata the canonical place for these?

```rust
use cargo_metadata::Target;

// https://github.com/rust-lang/cargo/blob/c24a09772c2c1cb315970dbc721f2a42d4515f21/src/cargo/util/machine_message.rs
#[derive(Deserialize, Debug)]
#[serde(tag = "reason", rename_all = "kebab-case")]
pub enum Message {
	CompilerArtifact {
		#[serde(flatten)]
		artifact: Artifact,
	},
	CompilerMessage {},
	BuildScriptExecuted {},
	#[serde(skip)]
	Unknown, // TODO https://github.com/serde-rs/serde/issues/912
}

#[derive(Deserialize, Debug)]
pub struct Artifact {
	pub package_id: String,
	pub target: Target,
	pub profile: ArtifactProfile,
	pub features: Vec<String>,
	pub filenames: Vec<PathBuf>,
	pub fresh: bool,
}

#[derive(Deserialize, Debug)]
pub struct ArtifactProfile {
	pub opt_level: String,
	pub debuginfo: Option<u32>,
	pub debug_assertions: bool,
	pub overflow_checks: bool,
	pub test: bool,
}
```

roblabla:
Fixes #45 

Most of the structs come directly from cargo or rustc. This is not ready yet, there are a few things that need to be done. Most importantly, the enums (mostly Message) need to have a catch-all variant so that it won't break if new variants are added. Also, I think we'd do well to change the names of some structs here and there.

Also, I need to add some tests and fix the doctest example.


roblabla:
For my sanity's sake, I'm currently working on fixing serde-rs/serde#912. This will allow turning Message into

```
enum Message {
    CompilerArtifact(Artifact),
    CompilerMessage(FromCompiler),
    BuildScript(BuildScriptExecution),
    #[doc(hidden)]
    #[serde(other)]
    Unknown
}
```

Blocking on https://github.com/serde-rs/serde/pull/1382

src/diagnostic.rs
```diff
@@ -0,0 +1,118 @@
+//! This module contains `Diagnostic` and the types/functions it uses for deserialization.
```
oli-obk: This file should be shared with rustfix in some manner

cc @killercup

roblabla:
Serde 1.0.79 is released, and with it #[serde(other)] is now available. I pushed a commit updating the serde dep to reflect that.

Question: Should all enums currently having a `DoNotMatchExhaustively` arm be switched to `#[serde(other)]`? There are only two as of this PR (not counting Message): `dependency::DependencyKind` and `diagnostic::Applicability`.

It's not a breaking change API-wise, as the arm is #[doc(hidden)]'d. Basically:

```rust
// Before
assert!(serde_json::from_str::<DependencyKind>(r#" "unknown" "#).is_error());

// After
assert_eq!(serde_json::from_str::<DependencyKind>(r#" "unknown" "#),
     Ok(DependencyKind::Unknown));
```

konstin:
I just tried the branch and I really like it. It makes a nice improvements over my earlier self-written structs, esp. with `parse_message_stream`.

One thing that would be useful is a `ToString` impl on `FromCompiler` using the `rendered` message. (I also wonder why `rendered` is an `Option<String>`. Is there a case were there is no rendered message?)

src/lib.rs
```diff
@@ -272,6 +302,95 @@ pub enum CargoOpt {
...
+    /// The workspace member this artifact belongs to
+    pub package_id: WorkspaceMember,
+    /// The target this artifact was compiled for
+    pub target: Target,
+    /// The profile this artifact was compiled with
+    pub profile: ArtifactProfile,
+    /// The enabled features for this artifact
+    pub features: Vec<String>,
+    /// The full paths to the generated artifacts
+    pub filenames: Vec<String>,
```
konstin: Might those be better as PathBuf than as String?

src/lib.rs
```diff
@@ -87,6 +87,33 @@
...
+//! ```
+//! # extern crate cargo_metadata;
+//! use std::process::{Stdio, Command};
+//! use cargo_metadata::Message;
+//! let command = Command::new("cargo")
+//!     .args(&["build", "--message-format=json"])
+//!     .stdout(Stdio::piped())
+//!     .spawn().unwrap();
+//!
+//! for message in cargo_metadata::parse_message_stream(command.stdout.unwrap()) {
```
konstin: Imo this example should `.take()` before unwrapping. Otherwise you can't wait for the command to finish afterwards.

src/lib.rs
konstin: Imho workspace_root should also be changed, but maybe in a different PR. @oli-obk What do you think?

mati865: @konstin Oliver is away until October.

konstin:
What would be required to get this merged and released on crates.io? I'm using those types in pyo3-pack and would like to release on crates.io without vendoring tricks.

roblabla:
I'll get around to addressing all the comments in the PR tomorrow (Using PathBuf, changing enums into `#[serde(other)]`, fixing the doctest, implementing ToString). Then I think it should be ready for merging?

src/diagnostic.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// A diagnostic message generated by rustc
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Diagnostic {
+    /// The error message of this diagnostic.
+    pub message: String,
+    /// The associated error code for this diagnostic
+    pub code: Option<DiagnosticCode>,
+    /// "error: internal compiler error", "error", "warning", "note", "help"
+    pub level: String,
```
epage: Is there a reason this isn't an enum?

epage:
I've forked this in my PR for escargot

Main differences
- I use `Cow` / `serde(borrowed)` everywhere
- I went ahead and switched to `Path`
- `Diagnostic::level` is an enum. 
- I have a `strict_unstable` feature that removes the `Unknown` variants and adds `serde(deny_unknown_fields)` that I use with my tests in the hope that it will help me identify when new fields are added

In addition, I went ahead and forked `WorkspaceMember` and `Target` rather than re-exporting `cargo-metadata`.  I find it curious that `WorkspaceMember` serializes to `raw: String` with parse functions rather than the components it makes up but I went ahead and kept that.

src/diagnostic.rs
epage: I guess the same could be said for `cargo-metadata`s existing `kind`, `crate_types`, `opt_levels`

src/lib.rs
```diff
@@ -339,3 +458,13 @@ pub fn metadata_run(manifest_path: Option<&Path>, deps: bool, features: Option<C
     let meta = serde_json::from_str(stdout)?;
     Ok(meta)
 }
+
+/// An iterator of Message.
+type MessageIterator<R> = serde_json::StreamDeserializer<'static, serde_json::de::IoRead<R>, Message>;
+
+// TODO: Should I use de::Read instead, to support deserializing from a slice?
```
oli-obk: If it's strictly more powerful to use `de::Read`, then by all means

src/lib.rs
```diff
@@ -339,3 +458,13 @@ pub fn metadata_run(manifest_path: Option<&Path>, deps: bool, features: Option<C
...
+/// Creates an iterator of Message from a Read outputting a stream of JSON
+/// messages. For usuage information, look at the top-level documentation.
```
oli-obk:
```suggestion
/// messages. For usage information, look at the top-level documentation.
```

src/diagnostic.rs
oli-obk: Yea, this was mostly chose for forward compatibility

src/lib.rs
roblabla: On investigating this a bit, de::Read isn't very convenient. It's implemented on serde_json::de::IoRead, serde_json::de::SliceRead, etc... and not on the actual libstd. Besides, I have a hard time envisioning under which situation would someone try to read an array of cargo messages from a byte array instead of an std::io::Read...

roblabla:
Superseded by #64

PR #56, opened 2018-12-03
Fix error when no categories are provided

Phlosioneer:
The error seems to be extremely rare, for some reason.

Conditions:
- rustc 0.24.0 (Stable)
- Linux (travis build, so I don't know which one exactly)
- cargo_metadata version 0.6.2
- Project is nested:
-- Parent project isn't published
-- Child project is published
-- cargo_metadata called from tests in the parent project
-- Parent project isn't a workspace
-- Parent project `Cargo.toml` is minimal; no `categories` entry.
- (Specific build log: https://travis-ci.org/budziq/rust-skeptic/jobs/462642234 from budziq/rust-skeptic#91)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -176,8 +176,10 @@ pub struct Package {
     /// Path containing the `Cargo.toml`
     pub manifest_path: String,
     /// Categories as given in the `Cargo.toml`
+    #[serde(default)]
     pub categories: Vec<String>,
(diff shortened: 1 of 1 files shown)


oli-obk:
Oh wow! Yea we can fix this backwards compatibly.

oli-obk:
published as 0.6.3

Phlosioneer:
Thanks for the prompt reply & publish.

PR #57, opened 2018-12-26
Add more fields

kornelski:
Current Cargo has more data. The `root` is especially important for working with workspaces.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -133,6 +133,9 @@ pub struct Metadata {
 pub struct Resolve {
     /// Nodes in a dependencies graph
     pub nodes: Vec<Node>,
+
+    /// The crate for which the metadata was read
+    pub root: Option<String>,
     #[doc(hidden)]
     #[serde(skip)]
     __do_not_match_exhaustively: (),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -145,6 +148,25 @@ pub struct Node {
     pub id: String,
     /// List of opaque identifiers for this node's dependencies
     pub dependencies: Vec<String>,
+
+    /// Dependencies in a structured format
+    #[serde(default)]
+    pub deps: Vec<NodeDep>,
+    /// Features provided by the crate
```
ehuss: I think this would be better to say that it is the set of features *enabled* on the crate.  The set of features available is something different.

src/lib.rs
```diff
@@ -145,6 +148,25 @@ pub struct Node {
     pub id: String,
     /// List of opaque identifiers for this node's dependencies
     pub dependencies: Vec<String>,
+
+    /// Dependencies in a structured format
```
ehuss: It might be helpful to explain the difference between this and `dependencies`.  In particular, `deps` handles renamed dependencies whereas `dependencies` does not.

src/lib.rs
```diff
@@ -145,6 +148,25 @@ pub struct Node {
...
+    pub features: Vec<String>,
+    #[doc(hidden)]
+    #[serde(skip)]
+    __do_not_match_exhaustively: (),
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+/// A dependency in a node
+pub struct NodeDep {
+    /// Crate name
```
ehuss: For renamed dependencies, this is the new (renamed) name, not the original name.

src/lib.rs
```diff
@@ -145,6 +148,25 @@ pub struct Node {
...
+    pub name: String,
+    /// Opaque unique identifier
```
ehuss: It might be helpful to call this a "Package ID".

oli-obk:
Thanks! Going through the PRs now and collecting all breaking changes for a new release

kornelski:
BTW, this is an incremental addition in structs with non-exhaustive field, so I expect it to be backwards-compatible (semver-minor).

PR #59, opened 2019-01-02
Unstringlytypify

### LINKED ISSUE #51: `WorkspaceMember` is not entirely correct.
In cargo's output, package id is opaque identifier. The proper way to extract package's name, version and stuff is to look it in the `packages` array by the id and inspect the fields. `WorkspaceMember` instead [splits the id](https://docs.rs/cargo_metadata/0.6.0/src/cargo_metadata/lib.rs.html#233-250). This works at the moment, but this is not guaranteed to work.

matklad:
closes #51

cc https://github.com/rust-lang/cargo/pull/6511

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,7 @@ use std::env;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::from_utf8;
+use std::fmt;
 
 pub use errors::{Error, ErrorKind, Result};
 pub use dependency::{Dependency, DependencyKind};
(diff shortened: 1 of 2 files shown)


tests/selftest.rs
```diff
@@ -39,21 +38,18 @@ fn metadata() {
...
-        let test_package_metadata: TestPackageMetadata = serde_json::from_value(value.clone())
-            .unwrap();
-        assert_eq!(test_package_metadata, TestPackageMetadata {
-            some_field: true,
-            other_field: "foo".into(),
-        });
-    }
+    let package_metadata = &metadata.packages[0].metadata.as_object()
+        .expect("package.metadata must be a table. \
+        NOTE: This test currently only works on the beta and nightly channel.");
```
ehuss: This line can probably be removed.

src/lib.rs
```diff
@@ -121,41 +121,71 @@ use std::env;
...
+/// It is possible to inspect the `repr` field, if the need arises, but its
+/// precise format is an implementation detail and is subject to change.
+///
+/// `Metadata` can be indexed by `PackageId`.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[serde(transparent)]
+pub struct PackageId {
+    /// The underlying string representation of id.
+    pub repr: String
+}
```
oli-obk: This move lost the convenience methods that existed on `WorkspaceMember`

oli-obk: Oh nevermind. I should have read the issue first

matklad: Yep, they are now basically `metadata[package_id].url` etc

PR #61, opened 2019-01-05
Add some more fields.

ehuss:
This is to help assist with stabilizing alternative registries.

- `Dependency::rename` — Stabilized in 1.31.
- `Dependency::registry` — Recently added in https://github.com/rust-lang/cargo/pull/6500
- `Dependency::target` — Add typed as suggested in https://github.com/oli-obk/cargo_metadata/pull/41#issuecomment-403742363.  Parsing this is rather complex (see https://github.com/rust-lang/cargo/blob/dcb4360c50574483dc96e8862afcb1e581c7b95c/src/cargo/util/cfg.rs).
- `Package::links` — Recently added in https://github.com/rust-lang/cargo/pull/6480

The `registry` field could use a URL type (with `url` and `url_serde` crates) if you want it to be more strictly typed (ala #59).


oli-obk:
Please make all of the new fields `#[serde(default)]` so it doesn't break using a newer `cargo_metadata` version for older compilers

src/dependency.rs
```diff
@@ -51,8 +52,29 @@ pub struct Dependency {
...
+    /// If None, the dependency is from crates.io.
+    pub registry: Option<String>,
     #[doc(hidden)]
     #[serde(skip)]
     __do_not_match_exhaustively: (),
 }
+
+/// A target platform.
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct Platform(String);
```
matklad: should this be `#[serde(transparent)]`?

ehuss: Hm, I don't understand the difference.  What does `transparent` do that the normal newtype struct doesn't do?

ehuss:
> Please make all of the new fields `#[serde(default)]` so it doesn't break using a newer `cargo_metadata` version for older compilers

I'm a little confused.  Isn't that the default behavior for `Option` types?  They are set to None if not present?

src/dependency.rs
matklad: For JSON there's no difference I think, but some data formats may include the name for unit-structs, [which is exposed for non-transparent wrappers](https://github.com/serde-rs/serde/blob/fea4e8e5b69bf44c051066aa9621f80144275992/serde/src/ser/mod.rs#L891)

oli-obk:
>  Isn't that the default behavior for Option types

I don't think so. But since you are mentioning it, a test would be great ;)

src/dependency.rs
ehuss: Ah. It also generates less code, which seems nice.

oli-obk:
Absolutely wonderful! Thanks

matklad:
@oli-obk I'd appreciate a release :wink:

oli-obk:
ups, done

PR #68, opened 2019-03-03
Make Package field source public

kolloch:
Well, I need it for a private project right now and I guess it was private by mistake to begin with.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -294,7 +294,9 @@ pub struct Package {
     pub authors: Vec<String>,
     /// An opaque identifier for a package
     pub id: PackageId,
-    source: Option<String>,
+    /// The source of the package, e.g.
+    /// crates.io or `None` for local projects.
+    pub source: Option<Source>,
     /// Description as given in the `Cargo.toml`
     pub description: Option<String>,
     /// List of dependencies of this particular package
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -294,7 +294,10 @@ pub struct Package {
     pub authors: Vec<String>,
     /// An opaque identifier for a package
     pub id: PackageId,
-    source: Option<String>,
+    /// The source of the package, e.g. 
+    /// "registry+https://github.com/rust-lang/crates.io-index"
+    /// or `None` for local projects.
```
ehuss: This may be more precisely stated that it is None for path dependencies or workspace members.

kolloch: Makes sesne!

src/lib.rs
```diff
@@ -294,7 +294,10 @@ pub struct Package {
     pub authors: Vec<String>,
     /// An opaque identifier for a package
     pub id: PackageId,
-    source: Option<String>,
+    /// The source of the package, e.g. 
+    /// "registry+https://github.com/rust-lang/crates.io-index"
+    /// or `None` for path dependencies or workspace members
+    pub source: Option<String>,
```
oli-obk: I think the reason it was private is that I was wondering whether we'd want to have a structured way to access this field. Can you create a newtype around `String`, keeping the `String` field private and only offering it via a `Display` implementation?

oli-obk:
ping @kolloch

src/lib.rs
kolloch: Not a bad idea but I fear slightly inconsistent. E.g. source in Dependency is still an `Option<String>`

src/lib.rs
```diff
@@ -360,6 +362,23 @@ pub struct Package {
     __do_not_match_exhaustively: (),
```
kolloch: BTW, I like that trick!

kolloch:
Ping, just in case that you missed that I updated this.

oli-obk:
oops yea. force pushes don't show up as email notifications.

oli-obk:
published in 0.7.4

PR #69, opened 2019-03-03
Correctly pass comma-separate features to cargo

kolloch:
Instead of wrapping them in '[' and ']'.

This change is untested!


oli-obk:
Oops, thanks!

PR #71, opened 2019-03-12
Add more customizable options to MetadataCommand.

### LINKED ISSUE #70: Add more flexibility to MetadataCommand
Particularly, I want the ability to run a nightly cargo with some `-Z` commands.  If you want this, when I implement it I'll submit a PR.

icefoxen:
Resolves #70.

Doesn't include any tests because I couldn't think of a good way to test it.  If you have any suggestions I'll happily add some.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -419,8 +419,9 @@ impl MetadataCommand {
     pub fn new() -> MetadataCommand {
         MetadataCommand::default()
     }
-    /// Path to `cargo` executable.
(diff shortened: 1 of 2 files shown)


oli-obk:
>  I couldn't think of a good way to test it.

One way would be to install multiple toolchains via rustup and manually switch between their `cargo`s. Seems somewhat overkill to me though. The impl is pretty straight forward, so there's not too much that needs testing anyway.

Thanks!

PR #72, opened 2019-03-13
add executable to artifact

matklad:
see https://github.com/rust-lang/cargo/blob/0e35bd8af0ec72d3225c4819b330b94628f0e9d0/src/cargo/util/machine_message.rs#L40


oli-obk:
published as 0.7.3

PR #73, opened 2019-03-13
Replace trim_left_matches with trim_start_matches

### LINKED ISSUE #65: Documentation uses deprecated trim_left_matches
`trim_left_matches` has been (or going to be soon?) deprecated in favor of `trim_start_matches`.
Perhaps the documentation should be updated to use the latter?

imp:
Now that 1.33.0 is out and `trim_left_matches()` deprecation has reached stable channel it may be a good time to update the example/doc test

Fix #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
     - rust: "nightly"
     - rust: "beta"
     - rust: "stable"
-    - rust: "1.27.0"
+    - rust: "1.30.0"
 
 script:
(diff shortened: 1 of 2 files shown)


oli-obk:
You'll need to bump the minimum supported version (in `travis.yml`) to the first rustc version that has `trim_start_matches`

imp:
Right, just pushed the fix

oli-obk:
Thanks. I'll include this in the next version bump

PR #78, opened 2019-03-29
Use failure based error instead of error_chain.

### LINKED ISSUE #67: Consider switching to a error type which is `Sync`
Consider switching to a error type which is `Sync` (so that it can be turned into a `Failure` error).

`Fail` requires `Send`+`Sync`+`'static` but error chain errors are not `Sync` (due to the way they handle backtraces if I remember correctly).

rustonaut:
Fix #67

Now Error is `'statis + Send + Sync` before it
was not `Sync` which could cause problems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT"
 readme = "README.md"
 
 [dependencies]
-error-chain = {version = "0.12.0", default-features = false}
+failure = "0.1.5"
 serde = "1.0.79"
 serde_derive = "1.0.79"
 serde_json = "1.0.1"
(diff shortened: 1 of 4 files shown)


rustonaut:
Sorry for taking so long, a lot of work came my way and I somehow forgot that I didn't yet made a pull-request.

rustonaut:
Also a Version bump (to 0.8) is still needed as the `Error` type changed.

oli-obk:
Super! Thanks

phil-opp:
@oli-obk Could you release a new version on crates.io that includes this PR?

PR #85, opened 2019-07-26
replace failure with just manually implementing an error

matklad:
We don't actually use any of failure's features anyway, and having less public dependencies is good for a library.

Here's what we lose from this:

```
λ cargo update
    Updating crates.io index
    Removing backtrace v0.3.33
    Removing backtrace-sys v0.1.31
    Removing cc v1.0.38
    Removing cfg-if v0.1.9
    Removing failure v0.1.5
    Removing failure_derive v0.1.5
    Removing rustc-demangle v0.1.15
    Removing synstructure v0.10.2
```

Note that technically this is a breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,6 @@ license = "MIT"
 readme = "README.md"
 
 [dependencies]
-failure = "0.1.5"
 serde = "1.0.79"
 serde_derive = "1.0.79"
 serde_json = "1.0.1"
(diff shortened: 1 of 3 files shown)


matklad:
:crab:  @oli-obk friendly ping :crab:

oli-obk:
Will this impact how other crates can use our errors?

matklad:
I don’t think so: there’s a blanket imp Fail for std::error::Error so those relying on Fail being implemented can continue to do so.

PR #87, opened 2019-09-08
package.readme should be Option<PathBuf> rather than Option<String>

imp:
Since the `Package::readme` field is public this would be a breaking change.
However, the ~aaec6f1~ 5448745  by itself should not break anything, while providing handy accessors, even without the updating (~3706bb2~ 66c3296) `Package::readme` field.


oli-obk:
Thanks, I'll check for other breaking changes we could do to improve the API and then publish a new breaking version bump

PR #88, opened 2019-10-10
Add `publish` and `doctest` fields.

ehuss:
`publish` added in 1.39 in https://github.com/rust-lang/cargo/pull/7354.
`doctest` added in 1.37 in https://github.com/rust-lang/cargo/pull/6953.


src/lib.rs
```diff
@@ -400,6 +406,11 @@ pub struct Target {
     /// Rust edition for this target
     #[serde(default = "edition_default")]
     pub edition: String,
+    /// Whether or not this target has doc tests enabled, and the target is
+    /// compatible with doc testing.
+    ///
+    /// This is `None` if running with a version of Cargo older than 1.37.
```
oli-obk: If that is the only condition making it `None`, I think I'd prefer it to default to `true` if missing

ehuss: Sure, I was a little concerned that it would be misleading when running on older versions, but it's probably not a big deal.

PR #89, opened 2019-10-12
make diagnostic types pub

JasperDeSutter:
To be able to fully use `parse_messages`, pub access is required for diagnostic types. Currently, it's not possible to match on `Diagnostic`'s `level` field because `DiagnosticLevel` is private.


oli-obk:
Thanks!

PR #93, opened 2019-11-30
Add BuildScript::out_dir

ehuss:
Added in https://github.com/rust-lang/cargo/pull/7622.

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -75,6 +75,11 @@ pub struct BuildScript {
     pub cfgs: Vec<PathBuf>,
     /// The environment variables to add to the compilation
     pub env: Vec<(String, String)>,
+    /// The `OUT_DIR` environment variable where this script places its output
+    ///
+    /// Added in Rust 1.41.
+    #[serde(default)]
+    pub out_dir: PathBuf,
     #[doc(hidden)]
     #[serde(skip)]
     __do_not_match_exhaustively: (),


src/messages.rs
```diff
@@ -75,6 +75,11 @@ pub struct BuildScript {
     pub cfgs: Vec<PathBuf>,
     /// The environment variables to add to the compilation
     pub env: Vec<(String, String)>,
+    /// The `OUT_DIR` environment variable where this script places its output
+    ///
+    /// Added in Rust 1.41.
+    #[serde(default)]
+    pub out_dir: String,
```
oli-obk: This could be a `PathBuf`, right?

ehuss: You're right, good catch.

src/messages.rs
```diff
@@ -75,6 +75,11 @@ pub struct BuildScript {
     pub cfgs: Vec<PathBuf>,
     /// The environment variables to add to the compilation
     pub env: Vec<(String, String)>,
+    /// The `OUT_DIR` environment variable where this script places its output
+    ///
+    /// Added in Rust 1.41.
+    #[serde(default)]
```
matklad, line 81: This caused a runtime panic in rust-analyzer(nothing to problematic, just a minor annoyance :) ) I guess this is an argument for using options in similar cases in the future.

https://github.com/rust-analyzer/rust-analyzer/issues/5125

oli-obk: oh.. I guess just using `default` for older compilers isn't making this backwards compatible. Thanks, I'll keep that in mind for the future.

PR #97, opened 2020-01-02
Typo fixed

guedou:
Thanks a lot for this crate! Here is a typo that I spotted while reading the documentation. [codepsell](https://github.com/codespell-project/codespell) did not find more typos.


oli-obk:
Thanks! I should set up spell check in CI some day

PR #100, opened 2020-02-19
src/errors: fix compiler warning

krisztian-kovacs:
Fix the following warning:

warning: trait objects without an explicit `dyn` are deprecated
  --> src/errors.rs:96:34
   |
96 |     fn source(&self) -> Option<&(::std::error::Error + 'static)> {
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn ::std::error::Error + 'static`
   |
   = note: `#[warn(bare_trait_objects)]` on by default

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -93,7 +93,7 @@ impl fmt::Display for Error {
 }
 
 impl ::std::error::Error for Error {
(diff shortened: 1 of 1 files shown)


oli-obk:
Thanks!

PR #101, opened 2020-02-19
src/lib: fix license_file()

krisztian-kovacs:
According to the documentation Package.license_file() returns the full path
to the license file, if present in metadata. However, in reality it
returned the name of the license file concatenated to the full path of the
Cargo.toml file.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -400,7 +400,7 @@ impl Package {
     pub fn license_file(&self) -> Option<PathBuf> {
         self.license_file
             .as_ref()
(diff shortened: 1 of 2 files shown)


oli-obk:
🤣oops, good catch, thanks!

tamird:
Apologies for necroing this. Can you help me understand why `self.manifest_path.parent().unwrap_or(&self.manifest_path)` is a reasonable thing to do? i.e. how can it happen that `self.manifest_path.parent()` returns None?

PR #105, opened 2020-03-31
Remove distracting info from documentation

matklad:
Listing three separate ways to parse command-line arguments in Rust is
not really helpful. If you are using `cargo metadata`, it's quite
likely that you are already a somewhat of an expert in Rust, and it's
rather unlikely that you literally want to just wrap it as is.

OTOH, the fact that most of the crate-level docs is occupied by
argument parsing makes finding the relevant info more difficult.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,6 @@ version = "0.9"
 [features]
 default = []
 
-[dev-dependencies]
-clap = "2.26.0"
-docopt = "1.0.2"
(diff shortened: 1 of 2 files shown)


matklad:
Sorry, this was bugging me for quite a while, and now I've finally succumbed to the irritation :)

oli-obk:
I honestly don't remember writing this, thanks

PR #109, opened 2020-04-23
Out of tree

ijackson:
Hi. I like to do builds out of tree, with the actual source code not writeable by the build process. This is achieved with cargo --manifest-path=..../Cargo.toml --target-dir=target.  This does not work right now if one of the parents of the source code directory has its own cargo config: the cargo run by the cargo_metadata tests picks up the wrong cargo config.

In this MR I propose to fix this by changing directory to $CARGO_HOME.  This is not ideal: it would be better to change back to the invocation directory, but the invocation directory is not currently available (https://github.com/rust-lang/cargo/issues/8148)

I have checked that "cargo test" still works in both in-tree and out-of-tree builds.

### DIFF
--- a/tests/selftest.rs
+++ b/tests/selftest.rs
@@ -5,8 +5,8 @@ extern crate serde_json;
 #[macro_use]
 extern crate serde_derive;
 
-use std::env::current_dir;
-use std::path::{Path, PathBuf};
+use std::env::{set_current_dir, self};
+use std::path::{Path, PathBuf, Component::CurDir};
 
 use semver::Version;
 
@@ -18,10 +18,37 @@ struct TestPackageMetadata {
     other_field: String,
 }
 
+fn chdir_find_manifest() -> (PathBuf, PathBuf) {
+    // 1. defuses any chdir that cargo has done
+    //    (see https://github.com/rust-lang/cargo/issues/8148)
+    //  2. returns the manifest path for passing to builder etc.
+
+    // This is idempotent (which is important, since cargo might
+    // call any subset of our test functions in whatever order)
+    // and also threadsafe.
+
+    let manifest_dir : PathBuf =
+        env::var_os("CARGO_MANIFEST_DIR")
+        .as_ref()
+        .map(|v| Path::new(v))
+        .unwrap_or(CurDir.as_ref())
+        .to_owned();
+
+    if let Some(cargo_home) = env::var_os("CARGO_HOME") {
+        set_current_dir(cargo_home)
+            .expect("chdir to undo cargo's chdir to manifest dir");
+        // Best we can until https://github.com/rust-lang/cargo/issues/8148
+    }
+
+    let mut manifest_path = manifest_dir.clone();
+    manifest_path.push("Cargo.toml");
+
+    (manifest_dir, manifest_path)
+}
+
 #[test]
 fn metadata() {
-    let manifest_dir = current_dir().unwrap();
-    let manifest_path = "Cargo.toml";
+    let (manifest_dir, manifest_path) = chdir_find_manifest();
     let metadata = MetadataCommand::new().no_deps()
         .manifest_path(&manifest_path).exec().unwrap();
 
@@ -66,8 +93,9 @@ fn metadata() {
 
 #[test]
 fn builder_interface() {
-    let manifest_dir = current_dir().unwrap();
-    let manifest_path = "Cargo.toml";
+    let (manifest_dir, manifest_path) = chdir_find_manifest();
+    let manifest_path : &str = manifest_path.to_str()
+        .expect("manifest path should be utf8");
     let _ = MetadataCommand::new()
         .manifest_path(manifest_path)
         .exec()
@@ -136,7 +164,7 @@ fn cargo_path() {
 #[test]
 fn metadata_deps() {
     std::env::set_var("CARGO_PROFILE", "3");
-    let manifest_path = "Cargo.toml";
+    let (_manifest_dir, manifest_path) = chdir_find_manifest();
     let metadata = MetadataCommand::new()
         .manifest_path(&manifest_path)
         .exec()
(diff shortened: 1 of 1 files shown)


oli-obk:
So this doesn't require any actual changes, you just want to make sure we don't regress this by adding a test?

ijackson:
Oliver Scherer writes ("Re: [oli-obk/cargo_metadata] Out of tree (#109)"):
> So this doesn't require any actual changes, you just want to make
> sure we don't regress this by adding a test?

Sorry, perhaps I was unclear.

The existing tests fail for me in my usual build environment because
the nested copy of cargo picks up an inappropriate config, by hunting
up from the source directory.  With my changes they pass.  People
doing in-tree builds should find they still pass - indeed, I tested
that and also Travis agrees :-).


Let me make it a bit more concrete.  If user alice does this:

   cd ~/builds
   mkdir cargo_metadata
   cd cargo_metadata
   cargo check --manifest-path=/home/bob/sources/cargo_metadata/Cargo.toml --target-dir=target

then alice's actual cargo invocation finds ~alice/.cargo/ and so uses
alice's cargo configuration.  But the tests inside selftest.rs
themselves run cargo again.  Effectively what happens is this:

 # cargo test does this
   [ run some builds ]
   cd ~bob/sources/cargo_metadata
   ~alice/builds/cargo_metadata/.../selftest
 # inside selftest.rs
   cargo metadata

That copy of cargo looks in its cwd and parent directories - ie it
looks in the following places for .cargo and its configuration:
   ~bob/sources/cargo_metadata
   ~bob/sources/
   ~bob/
It then finds ~bob/.cargo and uses bob's configuration.  This is
clearly wrong.  In my case ~bob/.cargo has a configuration which
doesn't work for users other than bob.  alice's cargo metadata
command bombs out and the test fails.

This MR fixes that by running the nested copy of cargo in $CARGO_HOME
rather than in the source directory.  So instead, we have this:

 # cargo test does this
   [ run some builds ]
   cd ~bob/sources/cargo_metadata
   ~alice/builds/cargo_metadata/.../selftest
 # inside selftest.rs
   cd ~alice/.cargo
   cargo metadata --manifest-path=/home/bob/sources/cargo_metadata/Cargo.toml

And then it works.  The nested cargo sees the config from
~alice/.cargo, which is correct, and uses the manifest file from
~bob/sources, which is also correct.  It doesn't write anything to the
cwd so this is all file.

It would be better for the cd to be back to
~alice/builds/cargo_metadata/ but we don't know that directory.
But ~alice/.cargo (ie $CARGO_HOME) at least makes the nested cargo
pick up alice's cargo config rather than bob's.


Sorry to go on at such length.  I hope this explains things...

Regards,
Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.

If I emailed you from an address @fyvzl.net or @evade.org.uk, that is
a private address which bypasses my fierce spamfilter.

oli-obk:
Ah wonderful. Can you add something to CI that will fail if we ever brick this? I feel like it's easy to break testing again for you in the future.

ijackson:
Oliver Scherer writes ("Re: [oli-obk/cargo_metadata] Out of tree (#109)"):
> Ah wonderful. Can you add something to CI that will fail if we ever brick this?
> I feel like it's easy to break testing again for you in the future.

Tricky.  I'm not very familiar with Travis.  But I can take a look.
Thanks for the encouragement, anyway!

Regards,
Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.

If I emailed you from an address @fyvzl.net or @evade.org.uk, that is
a private address which bypasses my fierce spamfilter.

ijackson:
Ian Jackson writes ("Re: [oli-obk/cargo_metadata] Out of tree (#109)"):
> Tricky.  I'm not very familiar with Travis.  But I can take a look.
> Thanks for the encouragement, anyway!

I have been playing with this and the problems seem deeper than I
thought.  In particular, I find that when I wrote a more thorough test
for this situation, some of the doctests fail.

The reason for this is that the doctests suffer from the same issue:
the examples show .manifest_path("./Cargo.toml").  But if you actually
write code that does that it won't work right in out-of-tree builds.

I am tempted to suggest that the right fix is for *MetadataCommand*
to chdir to CARGO_HOME.  By default, if no other .current_dir() is
supplied.  And presumably a note in the documentation.

The result would be that the existing doctests (and, presumably,
everyone's code) would be made to work correctly.  Ie, it seems to me
that we ought to provide an interface that does the right thing by
default.

What do you think ?

In the meantime I have closed the MR because my more thorough tests
seem to have revealed that my original solution was incomplete.

Regards,
Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.

If I emailed you from an address @fyvzl.net or @evade.org.uk, that is
a private address which bypasses my fierce spamfilter.

oli-obk:
phew... I'm not sure, and don't have the time to really dig into this. Do you think it will become better if your cargo issue is resolved?

ijackson:
Oliver Scherer writes ("Re: [oli-obk/cargo_metadata] Out of tree (#109)"):
> phew... I'm not sure, and don't have the time to really dig into this. Do you
> think it will become better if your cargo issue is resolved?

It seems to me that changing back to the cargo original invocation
directory, if cargo supplied that in an env var, is something that
would definitely be sensible to do in MetadataCommand by default.

It would only happen if the env var was set - so, only when
cargo_metadata was being used from something run from within cargo.
In that case running the nested cargo from the same directory as the
original cargo invocation is clearly more correct and more likely to
produce correct results.

I *think* this is true of changing to CARGO_HOME, too.  But I am less
sure of the implications of that.

Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.

If I emailed you from an address @fyvzl.net or @evade.org.uk, that is
a private address which bypasses my fierce spamfilter.

PR #110, opened 2020-04-23
Split command parse

ijackson:
Hi.

I have an application where I wanted to save the output from cargo metadata, which cargo_metadata produces internally.

The Metadata struct contains two useful pieces of functionality: the Command builder and the parser (which is built with serde of course).  I think the best approach is simply to expose these two pieces so they can be used separately.  The split is still not complete because `--format-version` would need special handling, but it's good enough for my purposes.

I have tested this with `cargo check` (my tests were run with this rebased on top of my out-of-tree tests chdir MR branch).  There are no separate new tests because both the new functions `cargo_command` and `parse` are called by `exec`, so already tested.  `parse` is trivial in any case and provided partly in order to document what is going on.

Although a FromStr impl would have been possible, I thought that probably wasn't desirable.

Regards,
Ian.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -488,6 +488,13 @@ impl MetadataCommand {
         Ok(cmd)
     }
 
+    /// Parses `cargo metadata` output.  `data` must have been
+    /// produced by a command built with `cargo_command`.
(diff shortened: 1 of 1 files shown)


oli-obk:
Thanks! This API extension makes sense

PR #111, opened 2020-04-28
Some improvements to MetadataCommand

sunshowers:
* Potentially reduce the number of clones required.
* Make the `exec` method `&self`.

(The second one breaks API compat I believe, hope that's ok!)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -496,7 +496,7 @@ impl MetadataCommand {
     }
 
     /// Runs configured `cargo metadata` and returns parsed `Metadata`.
-    pub fn exec(&mut self) -> Result<Metadata> {
+    pub fn exec(&self) -> Result<Metadata> {
         let mut cmd = self.cargo_command()?;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -496,7 +496,7 @@ impl MetadataCommand {
     }
 
     /// Runs configured `cargo metadata` and returns parsed `Metadata`.
-    pub fn exec(&mut self) -> Result<Metadata> {
+    pub fn exec(&self) -> Result<Metadata> {
```
oli-obk, line 499: I understand the motiviation for this, but if it's not strictly necessary I'd like to withold it until the next major version bump. Please remove it from the PR and open an issue for it.

oli-obk:
Thanks! Is there any pressure on your side to see this as a released version soon? Or are you ok with accumulating more changes over a while before releasing?

sunshowers:
No immediate pressure! Totally ok to wait a bit. Thanks for checking.

PR #112, opened 2020-05-04
 Correct BuildScript.cfgs to be String rather than PathBuf

robojumper:
As per the the cargo documentation which [says](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg):

> single identifier, or any arbitrary key/value pair [...] key should be a Rust identifier, the value should be a string

`String` should cover this.

This is a breaking change.

ref rust-analyzer/rust-analyzer#4296

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -72,7 +72,7 @@ pub struct BuildScript {
     /// The paths to search when resolving libs
     pub linked_paths: Vec<PathBuf>,
     /// Various `--cfg` flags to pass to the compiler
-    pub cfgs: Vec<PathBuf>,
+    pub cfgs: Vec<String>,
     /// The environment variables to add to the compilation
     pub env: Vec<(String, String)>,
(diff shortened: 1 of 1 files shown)


src/messages.rs
```diff
@@ -68,11 +68,11 @@ pub struct BuildScript {
     /// The package this build script execution belongs to
     pub package_id: PackageId,
     /// The libs to link
-    pub linked_libs: Vec<PathBuf>,
+    pub linked_libs: Vec<String>,
```
oli-obk, line 71: I think this is ok to be `PathBuf`, as it's either a name (so relative path) or a full path. It can't be a not-pathlike name afaict

robojumper: Changed back.

oli-obk:
Whoa that was fast :D Anyway, I just found the original addition, and it's likely an overzealous application of the discussion in https://github.com/oli-obk/cargo_metadata/pull/53#discussion_r219164079 that caused this.

Issue #79, opened 2019-04-08
Cannot pass both no-default-features and features


epage:
All, NoDefault, and Features are in an `enum`, making them mutually exclusive.  This would be fine except that they are then stored in `MetadataCommand` as a `Option<CargoOpt>`.  This prevents doing `--no-default-features --features "one-of-the-default"`.

epage:
I was also surprised that `Features` owns a `Vec<String>`.  Before looking at the code, I assumed `MetadataCommand` owned a `std::process::Command` and that `MetadataCommand` just forwarded calls to it.

oli-obk:
> I assumed MetadataCommand owned a std::process::Command and that MetadataCommand just forwarded calls to it.

That... seems very reasonable. Not sure why I wrote the code the way it is. You should be able to address this without touching the user-exposed API.

epage:
> That... seems very reasonable. Not sure why I wrote the code the way it is. You should be able to address this without touching the user-exposed API.

iirc you'd lose the ability to `Clone`.

epage:
If we want to aim for consistently wrapping `process::Command`, here is [escargot's feature functions](https://docs.rs/escargot/0.4.0/escargot/struct.CargoBuild.html#method.all_features).

ricobbe:
Looks like #118 addresses this problem, which is great.  Do you have an estimate of when this fix will be published to crates.io?

oli-obk:
oops, I published a new version with this, thanks for the ping

PR #118, opened 2020-07-08
Advanced feature configuration

mciantyre:
The PR addresses #79 while remaining API compatible. `MetadataCommand` now supports multiple calls to `features()`. We can now express

```bash
cargo metadata --no-default-features --features feat1
```

 as

```rust
MetadataCommand::new()
    .features(CargoOpt::NoDefaultFeatures)
    .features(CargoOpt::SomeFeatures(vec!["feat1".into()])
    // ...
```

See the tests for a demonstration.

Note that this is API compatible, but not behaviorally the same. This might break any user who relies on subsequent calls to `features()` that override previous calls to `features()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -415,7 +415,12 @@ pub struct MetadataCommand {
     manifest_path: Option<PathBuf>,
     current_dir: Option<PathBuf>,
     no_deps: bool,
-    features: Vec<CargoOpt>,
+    /// Collections of `CargoOpt::SomeFeatures(..)`
+    features: Vec<String>,
+    /// Latched `CargoOpt::AllFeatures`
+    all_features: bool,
+    /// Latched `CargoOpt::NoDefaultFeatures`
+    no_default_features: bool,
     other_options: Vec<String>,
 }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -478,12 +502,14 @@ impl MetadataCommand {
...
-        if let Some(features) = &self.features {
-            match features {
-                CargoOpt::AllFeatures => cmd.arg("--all-features"),
-                CargoOpt::NoDefaultFeatures => cmd.arg("--no-default-features"),
-                CargoOpt::SomeFeatures(ftrs) => cmd.arg("--features").arg(ftrs.join(",")),
-            };
+        if !self.features.is_empty() {
+            cmd.arg("--features").arg(self.features.join(","));
+        }
+        if self.all_features {
```
oli-obk, line 508: does `cargo metadata` really accept `--all-features` together with any of the other feature flags?

src/lib.rs
```diff
@@ -448,8 +453,27 @@ impl MetadataCommand {
...
+    /// # use cargo_metadata::{CargoOpt, MetadataCommand};
+    /// MetadataCommand::new()
+    ///     .features(CargoOpt::NoDefaultFeatures)
+    ///     .features(CargoOpt::SomeFeatures(vec!["feat1".into(), "feat2".into()]))
+    ///     .features(CargoOpt::SomeFeatures(vec!["feat3".into()]))
+    ///     // ...
+    ///     # ;
+    /// ```
+    ///
+    /// Unlike `cargo metadata`, which disallows multiple `--all-features` or `--no-default-features`
```
oli-obk, line 469: we could assert `!self.all_features` in the `AllFeatures` branch and do the same thing for `no_default_features`, too. Then the behaviour would be the same.

oli-obk:
I think the breaking change is ok, it was never expected to be used to override the old value.

src/lib.rs
mciantyre: It does. These are equivalent, since `--all-features` supersedes the other feature options:

```
cargo metadata --manifest-path tests/all/Cargo.toml --all-features --no-default-features --features feat2
cargo metadata --manifest-path tests/all/Cargo.toml --all-features
```

Added a test for this usage.

src/lib.rs
mciantyre: Sounds good. We'll replicate `cargo metadata` behaviors, and `panic!()` if calling `features(AllFeatures)` or `features(NoDefaultFeatures)` more than once.

oli-obk:
Thanks!

PR #120, opened 2020-07-19
Remove broken tests.

### LINKED ISSUE #119: It seems metadata() test doesn't consider user-wide `.cargo/config` settings.
Hello, I discovered one test is failing on my machine. 

- macOS 10.15.6 (19G73)
- cargo 1.44.1 (88ba85757 2020-06-11)
- rustc 1.44.1 (c7087fe00 2020-06-17)

Here's test result.

```
running 6 tests
test cargo_path ... ok
test error1 ... ok
test error2 ... ok
test metadata ... FAILED
test metadata_deps ... ok
test builder_interface ... ok

failures:

---- metadata stdout ----
thread 'metadata' panicked at 'assertion failed: `(left == right)`
  left: `"/Users/Dev/Workshop/aliens/cargo_metadata/target"`,
 right: `"/Users/Dev/.build"`', tests/selftest.rs:25:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    metadata

test result: FAILED. 5 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Here's source code for the broken part.

    assert_eq!(
        current_dir().unwrap().join("target"),
        Path::new(&metadata.target_directory)
    );

IMO, this is because I overridden target-dir using user-wide config file at `$HOME/.cargo/config`.

Here's my config file content.

    [build]
    target-dir = ".build"

It seems the test should consider this configuration or need to be updated.

I tried `cargo test --target-dir target`, but it still fails. I'm not sure what to do.

eonil:
This fixes #119 by removing broken test cases.

### DIFF
--- a/tests/selftest.rs
+++ b/tests/selftest.rs
@@ -6,7 +6,7 @@ extern crate serde_json;
 extern crate serde_derive;
 
 use std::env::current_dir;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 use semver::Version;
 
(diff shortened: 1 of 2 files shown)


oli-obk:
thanks!

Issue #121, opened 2020-07-28
Compile error in 0.11.0


Jake-Shadle:
Getting an extremely strange compile error, strange as in it seems it should happen all of the time, but only happens when this crate is a transitive dependency?

```
error[E0252]: the name `Deserialize` is defined multiple times
 --> /home/jake/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo_metadata-0.11.0/src/dependency.rs:5:20
  |
4 | use serde::{Deserialize, Deserializer};
  |             ----------- previous import of the macro `Deserialize` here
5 | use serde_derive::{Deserialize, Serialize};
  |                    ^^^^^^^^^^^--
  |                    |
  |                    `Deserialize` reimported here
  |                    help: remove unnecessary import
  |
  = note: `Deserialize` must be defined only once in the macro namespace of this module
```

This error seems to be completely valid, but doesn't appear when you just compile this crate alone or when it's a direct depdency, as far as I can tell.

I'll open up a PR to fix it.

oli-obk:
ooh... yea, this happens when some other crate in the dep graph enables the `derive` feature of `serde`

oli-obk:
I yanked 0.11 and published 0.11.1

Jake-Shadle:
Great, thanks!

PR #122, opened 2020-07-28
Fix duplicate import by simplifying imports

Jake-Shadle:
Resolves: #121


oli-obk:
Thanks! CI is failing out of other reasons, I'll fix that independently

PR #123, opened 2020-07-30
bump serde dependency to 1.0.107

Fabian-Gruenbichler:
as older versions can't compile cargo_metadata because of https://github.com/serde-rs/serde/issues/1804


oli-obk:
Thanks!

PR #128, opened 2020-09-20
Fix #127: Make messages hashable

### LINKED ISSUE #127: Make messages hashable
I'd like to filter messages with `Itertools::unique` and/or put them to hash table, but `Message` doesn't implement required traits.

alopatindev:


oli-obk:
Seems fine to me. I wonder if I should start putting some of these features behind feature gates

PR #129, opened 2020-09-24
Add the Metadata.root_package convenience method.

### LINKED ISSUE #126: Add Metadata.root_package() convenience method
Something like `fn root_package(&self) -> Option<&Package>`. This would make it much easier to access the root package for which the metadata was queried.

@oli-obk btw, thanks for the excellent project here. I'm happy to hack on this (it is dead simple).

thedodd:
closes #126

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,6 +139,14 @@ pub struct Metadata {
     __do_not_match_exhaustively: (),
 }
 
+impl Metadata {
+    /// Get the root package of this metadata instance.
+    pub fn root_package(&self) -> Option<&Package> {
+        let root = self.resolve.as_ref()?.root.as_ref()?;
+        self.packages.iter().find(|pkg| &pkg.id == root)
+    }
+}
+
 impl<'a> std::ops::Index<&'a PackageId> for Metadata {
     type Output = Package;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -139,6 +139,16 @@ pub struct Metadata {
     __do_not_match_exhaustively: (),
 }
 
+impl Metadata {
+    /// Get the root package of this metadata instance.
+    pub fn root_package(&self) -> Option<&Package> {
+        self.resolve
+            .as_ref()
+            .and_then(|res| res.root.as_ref())
+            .and_then(|root| self.packages.iter().find(|pkg| &pkg.id == root))
```
oli-obk, line 148: This could also use the `?` operator instead of `and_then`:

```suggestion
        let res = self.resolve.as_ref()?;
        let root = res.root.as_ref()?;
        self.packages.iter().find(|pkg| &pkg.id == root)
```

thedodd: Np. If you prefer that style, I'll do that.

oli-obk:
Thanks!

thedodd:
Thanks @oli-obk!

PR #130, opened 2020-10-05
Add workspace metadata field.

naerbnic:
This adds support for the new "metadata" field found at the workspace level.

Also, this eliminates the workspace and project metadata fields from being serialized if they are null.

### DIFF
--- a/tests/test_samples.rs
+++ b/tests/test_samples.rs
@@ -112,6 +112,7 @@ fn old_minimal() {
     );
     assert!(meta.resolve.is_none());
     assert_eq!(meta.workspace_root, PathBuf::from("/foo"));
+    assert_eq!(meta.workspace_metadata, serde_json::Value::Null);
     assert_eq!(meta.target_directory, PathBuf::from("/foo/target"));
 }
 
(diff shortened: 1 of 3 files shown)


tests/test_samples.rs
```diff
@@ -165,6 +176,14 @@ fn all_the_fields() {
         meta.workspace_root.file_name().unwrap(),
         PathBuf::from("all")
     );
+    assert_eq!(
+        serde_json::from_value::<WorkspaceMetadata>(meta.workspace_metadata).unwrap(),
```
oli-obk, line 180: this test only works on stable, but not on beta or nightly.

naerbnic: Had a bug where I was picking the wrong metadata object. Replaced with workspace.metadata path

naerbnic:
There is still a failing test, but this one is tricky

Apparently by introducing a workspace object, the workspace_members field picks up all of the packages below the the top-level "all" package. This can potentially be fixed by changing the test (checking that there are 8 members in meta.workspace_members and possibly checking their names). This seems like something worth exercising in a test somewhere, however I think this may also be dropping already tested implementation.

From the maintainers perspective, would it be better to fit the test to the data, to rework the test to not accidentally include the other packages, or to create a separate test fixture for this?

oli-obk:
I think the way you adjusted the test is totally fine

PR #133, opened 2020-10-14
Fix #132 Failed to parse compiler message

### LINKED ISSUE #132: Failed to parse compiler message
```
$ rustc --version
rustc 1.46.0 (04488afe3 2020-08-24)

$ cargo new parsing-bug && echo 'fn main() {format_xml::xml! {<A><A><A><A></A></A><A></A><A><A><A>true</A></A></A><A><A><A>0</A></A><A><A><A>{a}</A></A></A><A><A>"A"</A>if let Some(a)=(a){<A>{a}</A>}</A></A></A></A>};}' > parsing-bug/src/main.rs && cd parsing-bug && echo 'format_xml = "0.1.5"' >> Cargo.toml && cargo build --message-format=json-diagnostic-rendered-ansi > out.txt
     Created binary (application) `parsing-bug` package
    Updating crates.io index
   Compiling format_xml v0.1.5
   Compiling parsing-bug v0.1.0 (/home/al/tmp/parsing-bug)
error: could not compile `parsing-bug`.

To learn more, run the command again with --verbose.
```

```rust
fn main() {
    let s = include_str!("../../parsing-bug/out.txt");
    for message in cargo_metadata::Message::parse_stream(s.as_bytes()) {
        match message.unwrap() {
            cargo_metadata::Message::TextLine(_) => panic!("failed to parse"),
            _ => (),
        }
    }
}
```

alopatindev:


alopatindev:
Doesn't build with Rust 1.32.0

```
error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/psm-0.1.11/src/lib.rs:197:40
    |
197 |     let mut callback: MaybeUninit<F> = MaybeUninit::new(callback);
    |                                        ^^^^^^^^^^^^^^^^
```

@oli-obk what's the purpose of supporting the old compiler version? Thanks.

Cargo.toml
```diff
@@ -10,7 +10,8 @@ edition = "2018"
 
 [dependencies]
 serde = { version = "1.0.107", features = ["derive"] }
-serde_json = "1.0.1"
+serde_json = { version = "1.0.59", features = ["unbounded_depth"] }
+serde_stacker = "0.1"
```
oli-obk, line 14: is this import really needed?

oli-obk:
I think bumping the minimal supported Rust version to 1.36 is ok

Cargo.toml
alopatindev: At least that's what serde [documentation suggests](https://docs.rs/serde_json/1.0.59/serde_json/struct.Deserializer.html#method.disable_recursion_limit)

> You will want to provide some other way to protect against stack overflows, such as by wrapping your Deserializer in the dynamically growing stack adapter provided by the serde_stacker crate

I guess without it we're risking to run out of stack.

But in this particular case it works without `serde_stacker` anyway:

```patch
diff --git a/Cargo.toml b/Cargo.toml
index d028bec..642079e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,6 @@ edition = "2018"
 [dependencies]
 serde = { version = "1.0.107", features = ["derive"] }
 serde_json = { version = "1.0.59", features = ["unbounded_depth"] }
-serde_stacker = "0.1"

 [dependencies.semver]
 features = ["serde"]
diff --git a/src/messages.rs b/src/messages.rs
index 7884a00..9b4fceb 100644
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -147,8 +147,7 @@ impl<R: BufRead> Iterator for MessageIter<R> {
         let message = line.map(|it| {
             let mut deserializer = serde_json::Deserializer::from_str(&it);
             deserializer.disable_recursion_limit();
-            let deserializer = serde_stacker::Deserializer::new(&mut deserializer);
-            Message::deserialize(deserializer).unwrap_or(Message::TextLine(it))
+            Message::deserialize(&mut deserializer).unwrap_or(Message::TextLine(it))
         });
         Some(message)
     }
```

What do you think: should we drop the dependence or not?

alopatindev: I wasn't able to reproduce stack overflow, so I'll add the change. Feel free to revert it if you want.

alopatindev:
No need to update to 1.36 now 🎉

Cargo.toml
oli-obk: just adding a dependency will not help protect against stack overflows. I presume that you'd need to invoke something from that crate

alopatindev: > let deserializer = **serde_stacker**::Deserializer::new(&mut deserializer);

Yes, that was the usage from the first commit

oli-obk: :facepalm: sorry, I totally overlooked that

PR #142, opened 2020-11-22
Pin a version of semver-parser to keep compatibility with 1.32

jyn514:
This fixes the following build error:

```
error[E0658]: non-builtin inner attributes are unstable (see issue #54726)
 --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/semver-parser-0.10.1/src/generated.rs:4:1
  |
4 | #![rustfmt::skip]
  | ^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ edition = "2018"
 
 [dependencies]
 semver = { version = "0.11.0", features = ["serde"] }
+# Used to pin a version, so that semver-parser will compile on 1.32
+# Remove this when the MSRV is bumped or `semver` is updated.
+semver-parser = "=0.10.0"
 serde = { version = "1.0.107", features = ["derive"] }
 serde_json = { version = "1.0.59", features = ["unbounded_depth"] }
 


oli-obk:
that works, too I guess. Restarting CI to run with the other PR

kjeremy:
Unfortunately pinning semver-parser to 0.10.0 gives us a compile time regression that was fixed in 0.10.1: https://github.com/steveklabnik/semver-parser/pull/50

CC rust-analyzer/rust-analyzer/pull/6620

jyn514:
@kjeremy this should be fixed upstream in semver-parser then, otherwise cargo_metadata will have to raise its MSRV.

matklad:
I don't see a reason for `cargo metadata` to support 1.32. It's a pretty old Rust version, and cargo-metadata is not exactly a vocabulary crate.

matklad:
Argh, the situation around semver-parser  0.10.0 is annoying, sorry for missing an unintentional MSRV bump on that :(

That said, @jyn514 @oli-obk, I don't think merging this PR is/was a good idea, given how Cargo works :)

Pining requirements with `=0.10.0` constraint takes away the privilege of the downstream users to select the minor version they want. For example, if another crates requires `0.10.1`, then it won't be compatible with cargo-metadata. Or, if semver-parser releases 0.10.2, downstream crates won't be able to pick it up. To the first approximation, spelling `=` in Cargo.toml should be *never* done. 

For the MSRV problem, the appropriate solution is for the *consumers* of `cargo-metadata` who care about MSRV to:

* pin `semver-parser` to `0.10.0` in *their* Cargo.lock.
* submit a PR to `semver-parser` to fix its MSRV

oli-obk:
Since this is only a temporary situation, I'd prefer to just keep the perf regression and wait for the upstream fix to be merged. I do realize pinning a version is not great. 

> Or, if semver-parser releases 0.10.2, downstream crates won't be able to pick it up. 

I will immediately publish a new version of this crate once that happens and yank the intermediate state

matklad:
Sure, no big deal, we'll just avoid upgrading cargo-metadata on our side!

oli-obk:
Thanks!

kjeremy:
It looks like #150 passed so we can move forward.

Issue #125, opened 2020-09-10
Docs for cargo_metadata::Package are inaccurate


sbrocket:
https://docs.rs/cargo_metadata/0.11.2/cargo_metadata/struct.Package.html

A package is not the same as a crate (https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html); a package contains one or more crates. The thing that the Package struct refers to as "targets" are actually what are built into "crates" (though "targets" seems to be the correct Cargo term there).

PR #143, opened 2020-11-22
Update documentation

jyn514:
- Note that a package is not the same as a crate, which closes https://github.com/oli-obk/cargo_metadata/issues/125
- Switch to intra-doc links now that they're stable

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -57,7 +57,7 @@ pub struct Dependency {
     ///
     /// Use the [`Display`] trait to access the contents.
     ///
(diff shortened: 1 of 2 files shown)


jyn514:
One sec, I want to link to https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html.

oli-obk:
Won't intra-doc links require us to bump MSRV?

jyn514:
@oli-obk not for the build itself - the crate will compile fine, the links in the documentation will just be broken unless using a recent version of rustdoc.

oli-obk:
Ah, that's fine then, thanks!

Issue #94, opened 2019-11-30
Consider using Platform type.


ehuss:
The [`Platform`](https://github.com/oli-obk/cargo_metadata/blob/786a9457ff0f9dfd654beae6a83e49ad5b5c8c66/src/dependency.rs#L74-L83) type in cargo_metadata is currently an opaque string.  Recently, Cargo's `cfg()` parsing has been extracted into a separate crate: [cargo-platform](https://crates.io/crates/cargo-platform).  I'm filing this issue to let people know about it, and to see if it would be worthwhile to use its `Platform` type in cargo_metadata instead of just an opaque string.  This would provide a way for users to get structured access to the cfg expressions.

I'm not sure if it would actually be useful, though, since correlating cfg expressions with the compiler is a nontrivial task.  There is an example showing what needs to be done in [`matches.rs`](https://github.com/rust-lang/cargo/blob/c74ea1bc9692ad3cf8072e94c638b198da5174a9/crates/cargo-platform/examples/matches.rs).  The `--filter-platform` flag was added to `cargo metadata` as an alternate solution that makes Cargo do the heavy lifting. I think the filter flag probably covers most use cases people would have for inspecting the platform, but maybe there are still cases where it would be useful?  (I personally can't think of any.)

PR #144, opened 2020-11-22
Use `cargo_platform` instead of an opaque string

jyn514:
Closes https://github.com/oli-obk/cargo_metadata/issues/94.

I'm not sure how to test this - maybe since it's in a different crate it doesn't need tests? The existing tests at least make sure that it parses successfully, though.


oli-obk:
Oh, that's cool.

I think we can reasonably state that the next release will be a major bump, so I'm going to start merging your PRs

Issue #81, opened 2019-06-26
Have custom error for non-existent cargo executable


ThomasdenH:
I just found the cause of https://github.com/rustwasm/wasm-pack/issues/637. I thought the system might not have found the manifest file, but it turned out the error occurred because the cargo executable could not be found. `cargo` is in my path, but the invalid `cargo` environment variable took precedence.

Would it be possible to have a custom error variant that is returned when the executable doesn't exist?

PR #145, opened 2020-11-22
Don't return a Result from `cargo_command`

jyn514:
It can't ever fail.

Found while working on https://github.com/oli-obk/cargo_metadata/issues/81, I hope to have a fix for that soon.


jyn514:
(This is a breaking change).

PR #146, opened 2020-11-22
Give a better message on IO errors

jyn514:
Before: `error: No such file or directory (os error 2)`
After: ``error: failed to start `cargo metadata`: No such file or directory (os error 2)``

Closes https://github.com/oli-obk/cargo_metadata/issues/81 (it's not a custom error but it should be much more clear what's going on).

I'm not sure the difference between 'failed to create the process' and 'the process ran but exited with an error' is very clear. I don't know how to improve it, though.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -76,9 +76,9 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Error::CargoMetadata { stderr } => {
-                write!(f, "failed to run `cargo metadata`: {}", stderr.trim_end())
+                write!(f, "`cargo metadata` exited with an error: {}", stderr.trim_end())
             }
-            Error::Io(err) => write!(f, "{}", err),
+            Error::Io(err) => write!(f, "failed to start `cargo metadata`: {}", err),
             Error::Utf8(err) => write!(f, "cannot convert the stdout of `cargo metadata`: {}", err),
             Error::ErrUtf8(err) => {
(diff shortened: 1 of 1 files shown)


oli-obk:
maybe we should use something like `eyre` to just attach different messages to all `?` sites?

jyn514:
Well, I don't think that would actually help much here - eyre can reduce the boilerplate so we wouldn't need `impl fmt::Display for Error` but it wouldn't make the error messages more clear I think.

oli-obk:
With `eyre` we can use `.context()` on each `?` site, so we can give targetted messages even if the types are the same. With the current scheme, if it's an io-error, we report the same message, irrespective of where it came from.

jyn514:
I'm not really sure how to put my objections into words ... `cargo_metadata` is a library and I would expect it not to have opinions about error handling. If it switched to eyre, it would have to put all the eyre types in the API. If you wanted to use a different error handling library, you'd have to _both_ add another library on top and then figure out a way to convert from eyre to the new library. `thiserror` at least doesn't put the type in the API, but it's still opinionated in that it's easy to end up with two different error handling libraries. I don't want to say I'm *strongly* opposed to switching, but it's definitely not my first choice and I wouldn't do it myself.

If your concern is just that this change is trying to be too specific for all IO errors, I can instead add an error variant and use `.map_err`.

oli-obk:
> If your concern is just that this change is trying to be too specific for all IO errors, I can instead add an error variant and use .map_err.

Oh right :laughing: we have this option, too. I should have probably first checked if we even have multiple conversion sites for io errors... This PR seems fine to me then, the wording helps a lot I think, so I'll just merge it

PR #148, opened 2020-11-29
NOISSUE - Derive `Eq` and `Hash` traits for `DependencyKind`

jmcconnell26:
We ran into this in `cargo-geiger` when migrating from `cargo::core` to `cargo_metadata`, where we used the DependencyKind enum as keys in a `HashMap`.
We have implemented a temporary work around, using a privately declared enum, but it would be great to be able to use this enum.
Thanks,
Josh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-target
 Cargo.lock
+.idea/
+target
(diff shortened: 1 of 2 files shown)


oli-obk:
Thanks, seems totally fine to me to derive this since there's a concrete use case.
This will take a bit to hit crates.io since I'm planning another major bump soonish

PR #149, opened 2020-12-18
Add support for new path dep field

jonhoo:
This is pending for when https://github.com/rust-lang/cargo/pull/8994 eventually lands, and would be very useful for https://github.com/rust-lang/rustfmt/issues/4247#issuecomment-748240591.

### DIFF
--- a/tests/test_samples.rs
+++ b/tests/test_samples.rs
@@ -4,7 +4,7 @@ extern crate semver;
 extern crate serde_json;
 
 use cargo_metadata::{CargoOpt, DependencyKind, Metadata, MetadataCommand};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 #[test]
 fn old_minimal() {
@@ -169,6 +169,7 @@ fn all_the_fields() {
         // test added in 1.47
         // homepage added in 1.49
         // documentation added in 1.49
+        // path added in 1.50
         eprintln!("Skipping all_the_fields test, cargo {} is too old.", ver);
         return;
     }
@@ -224,6 +225,13 @@ fn all_the_fields() {
     assert_eq!(path_dep.source, None);
     assert_eq!(path_dep.kind, DependencyKind::Normal);
     assert_eq!(path_dep.req, semver::VersionReq::parse("*").unwrap());
+    assert_eq!(
+        path_dep
+            .path
+            .as_deref()
+            .map(|p| p.ends_with("path-dep/Cargo.toml")),
+        Some(true),
+    );
 
     all.dependencies
         .iter()
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -2,6 +2,7 @@
 
 use semver::VersionReq;
 use serde::{Deserialize, Deserializer, Serialize};
+use std::path::PathBuf;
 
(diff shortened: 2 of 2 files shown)


jonhoo:
https://github.com/rust-lang/cargo/pull/8994 was just accepted and is being merged. Will land in Rust 1.51 if all goes according to plan.

jonhoo:
Is it reasonable to merge this before 1.51 is released, given that it's documented that the field will be `None` until 1.51? I'm asking specifically because I think https://github.com/rust-lang/rustfmt/issues/4247 needs this to land in order to start taking advantage of the new field in nightly builds.

oli-obk:
Yes, seems fine to me to merge this, we just need to change the tests in CI to make sure they work right now

tests/test_samples.rs
```diff
@@ -224,6 +225,13 @@ fn all_the_fields() {
     assert_eq!(path_dep.source, None);
     assert_eq!(path_dep.kind, DependencyKind::Normal);
     assert_eq!(path_dep.req, semver::VersionReq::parse("*").unwrap());
+    assert_eq!(
+        path_dep
+            .path
+            .as_deref()
```
oli-obk, line 231: Please don't use `as_deref`, otherwise we need to bump MSRV

jonhoo: This is in the tests though, not in the main library? Will change anyway, but I wonder if we should exempt tests from being subject to MSRV.

jonhoo:
@oli-obk I updated the test to only test for a `path` in 1.51+

tests/test_samples.rs
```diff
@@ -224,6 +225,15 @@ fn all_the_fields() {
     assert_eq!(path_dep.source, None);
     assert_eq!(path_dep.kind, DependencyKind::Normal);
     assert_eq!(path_dep.req, semver::VersionReq::parse("*").unwrap());
+    if ver >= semver::Version::parse("1.51.0").unwrap() {
+        assert_eq!(
+            path_dep
+                .path
+                .as_ref()
+                .map(|p| p.ends_with("path-dep/Cargo.toml")),
```
ehuss, line 233: I don't think the path ends with `Cargo.toml`, you may want something more like this:

```rust
        assert_eq!(
            path_dep.path.as_ref().unwrap().file_name().unwrap().to_str().unwrap(),
            "path-dep",
        );
```

jonhoo: Ah, yes, you're right, I forgot to update the path following the change from `manifest_path` to `path` upstream. Will fix shortly!

jonhoo: Also, the reason I did not assert the full string was that I believe it will be an absolute path, not a relative path. Is that not the case?

ehuss: It's an absolute path.  The snippet I posted calls `file_name` just to get the last part.  You could also check if it ends with `tests/all/path-dep`.

jonhoo: Ah, missed the call to `file_name`. I think `ends_with("path-dep")` is probably true enough to what we want to test, without the potential of generating false-positives if the test directories get moved around. If you'd strongly prefer the `file_name()` or a longer prefix for `ends_with` though, I'd be happy to update the PR.

jonhoo:
I think this is now ready to merge?

oli-obk:
Jup! Thanks for the ping

PR #152, opened 2021-02-24
[RFC] use camino::Utf8PathBuf for returned paths

sunshowers:
[`camino`] is a library I just released which provides thin wrappers over Path
and PathBuf which guarantee that their contents are UTF-8. The readme at
https://crates.io/crates/camino explains the advantages of using this
library.

Given that:
* `cargo` only supports UTF-8 paths for cross-platform compatibility
  reasons
* `serde` only supports UTF-8 paths: https://docs.serde.rs/src/serde/ser/impls.rs.html#776-778
* JSON, as a format, only supports UTF-8 strings

all valid metadata instances are already forced to be UTF-8. This means
that `cargo_metadata` can switch over to using UTF-8 paths without any
loss of functionality.

What do you think?

[`camino`]: https://crates.io/crates/camino

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,6 +78,7 @@
 //! let output = command.wait().expect("Couldn't get cargo's exit status");
 //! ```
 
+use camino::Utf8PathBuf;
 use std::collections::HashMap;
 use std::env;
 use std::fmt;
@@ -85,6 +86,7 @@ use std::path::PathBuf;
 use std::process::Command;
 use std::str::from_utf8;
 
+pub use camino;
 pub use semver::Version;
 
 pub use dependency::{Dependency, DependencyKind};
@@ -138,9 +140,9 @@ pub struct Metadata {
     /// Dependencies graph
     pub resolve: Option<Resolve>,
     /// Workspace root
-    pub workspace_root: PathBuf,
+    pub workspace_root: Utf8PathBuf,
     /// Build directory
-    pub target_directory: PathBuf,
+    pub target_directory: Utf8PathBuf,
     /// The workspace-level metadata object. Null if non-existent.
     #[serde(rename = "metadata", default, skip_serializing_if = "is_null")]
     pub workspace_metadata: serde_json::Value,
@@ -272,13 +274,13 @@ pub struct Package {
     pub license: Option<String>,
     /// If the package is using a nonstandard license, this key may be specified instead of
     /// `license`, and must point to a file relative to the manifest.
-    pub license_file: Option<PathBuf>,
+    pub license_file: Option<Utf8PathBuf>,
     /// Targets provided by the crate (lib, bin, example, test, ...)
     pub targets: Vec<Target>,
     /// Features provided by the crate, mapped to the features required by that feature.
     pub features: HashMap<String, Vec<String>>,
     /// Path containing the `Cargo.toml`
-    pub manifest_path: PathBuf,
+    pub manifest_path: Utf8PathBuf,
     /// Categories as given in the `Cargo.toml`
     #[serde(default)]
     pub categories: Vec<String>,
@@ -286,7 +288,7 @@ pub struct Package {
     #[serde(default)]
     pub keywords: Vec<String>,
     /// Readme as given in the `Cargo.toml`
-    pub readme: Option<PathBuf>,
+    pub readme: Option<Utf8PathBuf>,
     /// Repository as given in the `Cargo.toml`
     // can't use `url::Url` because that requires a more recent stable compiler
     pub repository: Option<String>,
@@ -344,7 +346,7 @@ pub struct Package {
 
 impl Package {
     /// Full path to the license file if one is present in the manifest
-    pub fn license_file(&self) -> Option<PathBuf> {
+    pub fn license_file(&self) -> Option<Utf8PathBuf> {
         self.license_file.as_ref().map(|file| {
             self.manifest_path
                 .parent()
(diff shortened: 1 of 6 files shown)


sunshowers:
Ah I guess one issue is that cargo_metadata supports Rust 1.32 and above while camino only supports Rust 1.39 or above (I picked Debian stable minus 2 versions). Looks like camino can support 1.34 without any changes -- I can release an update promising support for 1.34 if you think camino is useful and are willing to bump the MSRV to that.

src/lib.rs
```diff
@@ -78,6 +78,7 @@
 //! let output = command.wait().expect("Couldn't get cargo's exit status");
 //! ```
 
+use camino::Utf8PathBuf;
```
oli-obk, line 81: Please publicly reexport `camino` so that crates depending on `cargo_metadata` don't have to add a dependency on camino themselves (and potentially run into versioning issues in the future).

oli-obk (commented):
I'm totally on board with bumping to 1.34. I just have CI gate on our MSRV so we only bump consciously, not to actually stay back on a low MSRV.

src/lib.rs
sunshowers: Done, thanks.

sunshowers:
Updated the MSRV in CI to 1.34, and re-exported camino. Thanks for the review!

oli-obk:
\o/ CI is happy

I'll look over the open issues to see if anything else should go into the major bump and then I'll publish a new version. Since I want to give ppl the possibility to give feedback on these things, I'll wait until after the weekend.

oli-obk:
Published 0.13 with this change and some other (technically breaking) changes. We're at MSRV 1.40 now

matklad:
urgh, I am not sure the benefits of using `camio` apply to `cargo metadata`. At the same time, there's one strong and one weak drawback.

Not using a standard library type is a strong drawback, as now every use of `cargo metadata` needs to wonder "wait, what  is Utf8PathBuf?"

Adding an extra dependency is a weak drawback, as there are some cargo projects that care about keeping the count of dependencies as low as possible.

sunshowers:
Yeah, hmm. I'm not really sure what to do here. I do personally think camino's model is superior (and it's easy to convert back to a Path/PathBuf for compatibility). A feature would be the wrong way to do this because of unification issues. A type param seems excessive.

We could back this change out completely but in that case I would probably end up forking cargo_metadata for my own projects, which would suck.

matklad:
To be clear, I have no super strong feelings. I am also considering forking `cargo-metadata` for rust-analyzer anyway, as we are a heavy user, and it seems cheaper to build rather that by.

That being said, I don't think that `camio`  model is universally supperior. In rust-analyzer, we have our own module for working with paths: https://github.com/rust-analyzer/rust-analyzer/blob/c17f2bf2a27798858ef2e3012ca28295aed46efa/crates/paths/src/lib.rs. For us, it's important to distinguish relative vs absolute path, while we don't particularly care about encodings. In our context, `camio` adds an unnecessary layer which we need to peel to get `PathBuf`.

matklad:
To clarify, I do agree, that, *in principle*, Utf8Path model is better. But, because it is a *non-standard* model, this benefit needs to be weighted against the costs, and the net turns out to be negative in rust-analyzer's case.

sunshowers:
I really don't mean to upset anyone by doing this, so please feel free to revert this if necessary.

Agreed about rust-analyzer and other projects wanting their own path types, that's totally reasonable. (One possibility might be to wrap the camino types rather than the std types.)

oli-obk:
> I am also considering forking `cargo-metadata` for rust-analyzer anyway, as we are a heavy user, and it seems cheaper to build rather that buy :)

I would very much like to avoid that.

What comes to mind immediately would be to have a `cargo_metadata_core` that is generic over everything that anyone has opinions on. The default `cargo_metadata` would then go for convenience, but the heavier users would not have to reimplement everything from scratch and can instead reuse the datastructures. I would imagine that we'd not even have the `CargoMetadata` struct in the core crate, so you need to plug your own serde json deserializer.

sunshowers:
That seems like a decent idea, and I'm happy to put in the work for it if it'll prevent a fork. (Wondering if zero-copy deserialization through borrowed types can fit into the same framework as well.)

matklad:
>I would very much like to avoid that.

I personally don't find forking (or, rather, vendoring of only the relevant bits) to problematic in my context. In IntelliJ Rust, we had to maintain our own metadata. Empirically, the cost of maintaining [those 500 lines](https://github.com/intellij-rust/intellij-rust/blob/9e95fe452aba92b25b16bfd0f7ffac40e1ae3d85/src/main/kotlin/org/rust/cargo/toolchain/impl/CargoMetadata.kt) in IntelliJ Rust was lower than the cost of maintaining `cargo_metadata` dependency in `rust-analyzer`. 

I def wold prefer a bit of copy-pasete over using something which is generic over proc-macro )

That being said, there's long been a desire to exctract just the data definitions to share them with cargo. That might be a promising avenue.

ehuss:
I ran into this with a project trying to update.  I too think it is a little awkward to use non-standard types.  

Maybe I'm blind, but is there a way to convert a `Utf8PathBuf` to a `std::path::Path`?  It is common for API's to take a `Path`, and it seems like all uses will now need to do some kind of conversion?  That seems like it comes at a cost of reallocating all the paths, and extra code to do the conversion.

ehuss:
Ah, indeed I am blind. 😜   Thanks!  

I don't think it is feasible to use generics everywhere.  Rust does not handle those well, particularly if they get instantiated multiple times.  And internal APIs and small functions that are intended for basic and simple use would be hindered by them. I agree that outward-facing API's benefit from generics, but I don't think they should be used everywhere.  It's a bit unfortunate to require the explicit conversion, but I can deal with it.

Issue #156, opened 2021-04-14
impl Display for DependencyKind


jyn514:
I would like to write this code:
```rust
/// Convert dependencies into Vec<(String, String, String)>
fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)> {
    pkg.dependencies
        .iter()
        .map(|dependency| {
            let name = dependency.name.clone();
            let version = dependency.req.to_string();
            let kind = dependency.kind.to_string();

            (name, version, kind)
        })
        .collect()
}
```
but it doesn't compile:
```
error[E0599]: the method `to_string` exists for enum `DependencyKind`, but its trait bounds were not satisfied
   --> src/db/add_package.rs:206:40
    |
206 |             let kind = dependency.kind.to_string();
    |                                        ^^^^^^^^^ method cannot be called on `DependencyKind` due to unsatisfied trait bounds
    | 
   ::: /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/cargo_metadata-0.12.3/src/dependency.rs:9:1
    |
9   | pub enum DependencyKind {
    | -----------------------
    | |
    | doesn't satisfy `DependencyKind: ToString`
    | doesn't satisfy `DependencyKind: std::fmt::Display`
    |
    = note: the following trait bounds were not satisfied:
            `DependencyKind: std::fmt::Display`
            which is required by `DependencyKind: ToString`
```
I see that DependencyKind implements Serialize - maybe I'm supposed to use that instead?

oli-obk:
if there's a reasonable display impl, go right ahead and add it. If that re-uses serialiye even better

PR #159, opened 2021-04-22
Use serde(default) instead of custom deserialization

jyn514:
Found while working on https://github.com/oli-obk/cargo_metadata/issues/156.


jyn514:
Hmm, it looks like this breaks because cargo uses `null` instead of omitting the key altogether. Not sure if there's an easy way to fix that.

PR #172, opened 2022-02-13
implement Eq for Package

MarcoIeni:


oli-obk:
Are you interested in plain equality or will you need a diff anyway? In the latter case I would recommend serializing to json and using a json diff crate or tool

MarcoIeni:
For my use case I was just interested in equality. Anyway, apart from my use case, I submitted the PR because it's also suggested in the [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits).
Do you see any cons in implementing equality?

oli-obk:
published as 0.14.2

PR #174, opened 2022-04-17
Small changes to improve the overall crate quality

morr0ne:
see #175

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
           - rust: stable
           - rust: beta
           - rust: nightly
-          - rust: 1.40.0
+          - rust: 1.42.0
     steps:
     - uses: actions/checkout@v2
     - name: Install rust
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ description = "structured access to the output of `cargo metadata`"
 license = "MIT"
 readme = "README.md"
 edition = "2018"
+rust-version = "1.42.0"
 
 [dependencies]
 camino = { version = "1.0.7", features = ["serde1"] }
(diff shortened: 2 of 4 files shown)


oli-obk:
Yea, I think a new breaking change is overdue, there's probably more things we'll want to do, too. I'll collect them for a while on master and release when we don't have anything left we'll want to break. 

Please unbump the edition and adjust the tests for the msrv

morr0ne:
I think the best course of action is to eventually update the edition to avoid potential future problems. I reverted it back to the 2018 edition. There should be no conflicts now.

morr0ne:
I believe it might be beneficial to also specify a rust-version in the manifest file (I believe the current msrv is 1.40.0). According to the [docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) the field would just be ignored on older rust version

oli-obk:
I agree with both bumping the edition at some point and with adding the rust version to the manifest now

oli-obk:
With the use of the matches macro we'll need to bump to 1.42 in this PR

morr0ne:
I've been using rust for just 2 years so I never realized how relatively new the matches! macros is 😅. Should I revert it back? Using the macro is more ergonomic and readable but probably doesn't warrant a msrv bump.

oli-obk:
I think the bump is definitely ok. It's 18 versions ago, so 27 months. I like having a bit over a year, even if the rust survey suggests that two versions is enough 🤣

morr0ne:
Fair enough I guess lol. I actually never had to deal with msrv problems, I just use the latest nightly version that provides all the components. I'll update it as soon as I can. (Or possibly try from my phone 😬)

morr0ne:
I managed to make the necessary edits from my phone. It should be good to go now.

PR #176, opened 2022-04-24
add Package::is_publishable method

MarcoIeni:
I needed this tiny method in my code, so I opened a PR to add it here.
I already implemented this with a trait in my codebase, so feel free to close this PR if it is out of scope for this crate :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,6 +369,19 @@ impl Package {
             .as_ref()
             .map(|file| self.manifest_path.join(file))
     }
+
+    /// Return true if the field `publish` in Cargo.toml:
+    /// - is not `[]`
+    /// - is not `false`
+    pub fn is_publishable(&self) -> bool {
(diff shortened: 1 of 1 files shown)


oli-obk:
Can you elaborate on your use case? Why/when would it be irrelevant which registry the crate can be published to?

MarcoIeni:
In [release-plz](https://github.com/MarcoIeni/release-plz) I only update the publishable packages, because:
- it doesn't make sense to bump the version of packages which can't be published
- to determine the next version number, release-plz compares the local package with the one present in the cargo registry (it clones the published package and compares the two directories).

So I use this function to filter the workspace packages.

oli-obk:
I've thought a bit about this and I think the function is too niche to be included here. It hides the fact that it may only be publishable on specific registries, which is fine for your use case, but I'd like to keep it explicit for everyone else.

MarcoIeni:
Sure, no problem!

PR #177, opened 2022-04-24
add Metadata::workspace_packages method

MarcoIeni:
Inspired by [this](https://github.com/killercup/cargo-edit/blob/8a14de08c204e77374187f2d9348bae497803909/src/metadata.rs#L26) cargo-edit method, I thought it might be nice to have this functionality in this crate, directly.


oli-obk:
Makes sense!

Issue #179, opened 2022-06-22
Re-exporting `semver`?


8573:
Like [`camino`](https://lib.rs/crates/camino), [`semver`](https://lib.rs/crates/semver) is in `cargo_metadata`'s API, but, [unlike with `camino`](https://github.com/oli-obk/cargo_metadata/issues/171), `cargo_metadata` does not re-export the version of `semver` that it uses.  I suggest that `cargo_metadata` also re-export `semver` so that users can be sure of using the same version `cargo_metadata` uses.

I could file a PR, but
```diff
 pub use camino;
-pub use semver::{Version, VersionReq};
+pub use semver::{self, Version, VersionReq};
```
is such a simple change that I figure handling a PR would be more work for you. :-)

<small>(In case it matters, I dedicate the above patch under CC0-1.0.)</small>

UebelAndre:
https://github.com/oli-obk/cargo_metadata/commit/1b4aea8dee94652c4f27a5e2f4e0dc07d3da6e55 seems to be a followup change.

PR #180, opened 2022-06-22
Re-export semver::Prerelease and BuildMetadata

cecton:
This makes it easier for the user of cargo_metadata to manipulate the
Version without having to add semver to their dependencies.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,7 +89,7 @@ use std::process::Command;
 use std::str::from_utf8;
 
 pub use camino;
-pub use semver::{Version, VersionReq};
+pub use semver::{BuildMetadata, Prerelease, Version, VersionReq};
 
 pub use dependency::{Dependency, DependencyKind};
 use diagnostic::Diagnostic;


cecton:
wow thanks that was super fast!

oli-obk:
Nice! thanks for the tip. Using that addon now, too

jyn514:
Hmm, I wonder if we should just re-export the whole semver crate (https://github.com/oli-obk/cargo_metadata/issues/179) ?

cecton:
> Hmm, I wonder if we should just re-export the whole semver crate (#179) ?

tbh... on [xtask-wasm](https://docs.rs/xtask-wasm/latest/xtask_wasm/) and [xtask-watch](https://docs.rs/xtask-watch/latest/xtask_watch/) we did not bother. We re-exported the entire crates.

One good point when doing that is that it makes it clear that the structs belong to that particular crate and not to us.

On the downside, it's harder to import.

(Also it's a breaking API change if you change it now because it is still part of your public API)

(Note also that cargo_metadata already re-export the entire camino crate)

8573:
A point in favor of re-exporting the whole crate is that there are more items from it that are in the API — `Comparator` and `Op` were my main motivations.

PR #183, opened 2022-06-22
define Edition using a newtype over strings

### LINKED ISSUE #182: Edition should be a newtypes over a string for forward compatibility
Hi there — noticed an issue with the way the new enum Edition is defined.

Currently, Edition is a plain enum. This means that once the 2024 edition is available, the current version of cargo_metadata will no longer be able to parse it.

Instead, Edition should be defined as a newtype over a `Cow<'static, str>`, with the current options being associated constants. See https://sunshowers.io/posts/open-closed-universes/#many-strings.

sunshowers:
Previously, `Edition` was a plain enum. This means that once the 2024
edition is available, cargo_metadata would no longer be able to parse
it.

Instead, define `Edition` as a newtype over strings, to enable future
expandability while still being able to conveniently compare against
currently-known editions.

Fixes #182.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,7 @@
 use camino::Utf8PathBuf;
 #[cfg(feature = "builder")]
 use derive_builder::Builder;
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::env;
 use std::fmt;
@@ -450,19 +451,30 @@ pub struct Target {
     pub doc: bool,
 }
 
+/// The Rust edition
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
-#[non_exhaustive]
-/// The rust edition
-pub enum Edition {
-    /// Edition 2015
-    #[serde(rename = "2015")]
-    E2015,
-    /// Edition 2018
-    #[serde(rename = "2018")]
-    E2018,
-    /// Edition 2021
-    #[serde(rename = "2021")]
-    E2021,
+#[serde(transparent)]
+pub struct Edition(pub Cow<'static, str>);
+
+impl Edition {
+    /// Creates a new `Edition`.
+    pub fn new(edition: impl Into<Cow<'static, str>>) -> Self {
+        Self(edition.into())
+    }
+
+    /// Creates a new `Edition` from a static string.
+    pub const fn new_const(edition: &'static str) -> Self {
+        Self(Cow::Borrowed(edition))
+    }
+
+    /// Edition 2015.
+    pub const E2015: Self = Self::new_const("2015");
+
+    /// Edition 2018.
+    pub const E2018: Self = Self::new_const("2018");
+
+    /// Edition 2021.
+    pub const E2021: Self = Self::new_const("2021");
 }
 
 impl Default for Edition {


oli-obk:
neat trick!

morr0ne:
I kinda have to disagree with this change, the reason I made it an an enum was so it would be easy to match against it. Also this way there is no way to check against invalid edition since any string is accepted meaning one could easily write the following code:
```rust
// An over simplified example but you can see how this can allow bad input
let edition = Edition::new("Rushia");
```
And ending up with a completely invalid edition, furthermore it also forces matching against a Cow<str> instead of a simple enum discriminant.

I think a better solution would be something in the middle that utilizes the best of both word:
```rust
pub enum Edition {
    #[serde(rename = "2015")]
    E2015,
    #[serde(rename = "2018")]
    E2018,
    #[serde(rename = "2021")]
    E2021,
    Unknown(Cow<'static, str>),
}
```
This allows to properly match on the known edition but is still future proof, even in the unlikely case that the next edition is named `ferris`.

This is not even a breaking change and could easily be released under version `0.15.1` as the old enum was marked as `non_exhaustive`

sunshowers:
The unknown variant approach has more fundamental issues unfortunately:  https://sunshowers.io/posts/open-closed-universes/#a-pattern-to-avoid-unknown-or-other-variants

sunshowers:
I think for the invalid version check we can add a method returning a list of all known valid versions.

morr0ne:
> The unknown variant approach has more fundamental issues unfortunately: https://sunshowers.io/posts/open-closed-universes/#a-pattern-to-avoid-unknown-or-other-variants

I'm not sure how this applies in most cases, Sure there is no easy way to compare them but why would you? This types are being constructed by serde so Edition::Unknown("2018") will never happen since it will always be parsed as Edition::E2018.

Also it could probably be implemented by manually implement Eq/PartialEq on the Edition enum.

> I think for the invalid version check we can add a method returning a list of all known valid versions.
> 
> Matching based on editions is simply:
> 
> ```rust
> if package.edition == Edition::E2015 {
>     // ...
> } else if package.edition == Edition::E2018 {
>     // ...
> } else ...
> ```

Considering that we have actual match statement that isn't very ergonomic. Furthermore it doesn't really make much sense since the edition field will rarely ever not be one of those three otherwise it would be a bug on cargo side.

sunshowers:
I'm not sure what you mean. The entire point of this is to allow the current version of cargo_metadata to handle future editiona like 2024, etc. So you always have to handle an unknown case in your code.

morr0ne:
Regarding your blog post I believe that is as easy to fix as manually implementing Eq/PartialEq the following way:
```rust
use std::borrow::Cow;

pub enum Edition {
    E2015,
    E2018,
    E2021,
    Unknown(Cow<'static, str>),
}

impl PartialEq for Edition {
    fn eq(&self, other: &Self) -> bool {
        match (self, other) {
            (Self::Unknown(left), Self::Unknown(right)) => left == right,
            (Self::E2015, Self::Unknown(unknown)) => unknown == "2015",
            (Self::E2018, Self::Unknown(unknown)) => unknown == "2018",
            (Self::E2021, Self::Unknown(unknown)) => unknown == "2021",
            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
        }
    }
}

impl Eq for Edition {}
```
This check for equally in cases like:
```rust 
Edition::E2018 == Edition::Unknown("2018")
```



> I'm not sure what you mean. The entire point of this is to allow the current version of cargo_metadata to handle future editiona like 2024, etc. So you always have to handle an unknown case in your code.

The point is being able to match against edition knowing and knowing it is correct, if I match against my version I can easily handle all cases even if an eventual 2024 edition comes out and this crate isn't update, but with your version I cannot be sure I have have to match against a possibly wrong value. It's not much or why do it, is more of a why not. There are no downside since its still future proof but it's match more ergonomic on the library consumer side.

sunshowers:
While the Hash and Eq impls look ok to me, we may want to add Ord in the future (e.g. to store editions as keys in a BTreeMap) which becomes harder to define in a stable way. It just adds epicycles of complexity that just don't exist with the many-strings approach.

With the many-strings approach, you don't compare with unknown versions typically—you simply compare against all versions known to the library and defined as associated constants, plus any versions that you as the consumer know about but the library doesn't.

Simply put, enums are just not the right tool to model fully open universes like the edition field. This isn't just my opinion; I've had conversations with a number of other people who agree with this (and in fact it was a coworker who first mentioned this problem to me).

morr0ne:
I think that's the problem which we don't agree with, the edition field isn't an 
> open universes
It can literally only be this values, 2015, 2018 and 2021. That's it, there are no other possible values, any other value is wrong. Even if we want future proofing the next edition won't be released for another 2 and half years. By then we'd have plenty of time to simply add a single line to support edition 2024. Both this implemntation and mine over at #185 are probably quite useless. I think my original version over on #174 is more that enough. After all we do want to throw an error if we have an unexpected value.

sunshowers:
As a consumer of cargo_metadata who publishes a Rust dev tool as a binary crate, I promise to my consumers that old versions of my binary will work with future versions of user code 😊

morr0ne:
While I think it's a bit unreasonable I do understand the concern but a middle ground must be found. If Edition ends up being just a fancy wrapper around a String then we might as well just leave it as a string and match on it this way:
```rust
match edition {
"2018" => todo!()
"2021" => todo!()
// etc, you get the idea
}
```
There must be a way to satisfy both our needs, otherwise it's actually better to just leave it as a String and move on without the added complexity. The other cases need to be handled regardless if we stuff everything in a String or us an enum with an Unknown variant attached. I think there might be some way to make #185 work for both of us.
One important use case for me that I haven't highlighted here is that in very limited enviroments (like embeded boards running linux) I found that matching against an enum discriminant leads to significantly more optimized and faster code than dealing with a `Cow<'static, str>` which might end up triggering allocations, having to store the static str somewhere and in either case less optimized code paths when matching.

This is actually one of the reasons that lead me to my original implementation in #174

oli-obk:
We could make the field an integer and validate that it is larger than 2015 and in 3 year steps from then on.

sunshowers:
Making it an integer sounds fine to me :) thanks for the discussion, sorry it got a little acrimonious. I have a lot of sympathy for the embedded use case—hopefully switching to an integer will solve it.

morr0ne:
> We could make the field an integer and validate that it is larger than 2015 and in 3 year steps from then on.

Well in that case we might as well just have an `Edition` enum that includes more editions that the current ones, something like the following: 
```rust
enum Edition {
    E2015,
    E2018,
    E2021,
    E2024,
    E2027,
}
```
I think that ought to be more than enough, and if it isn't at some point we can just add another variant and mark it as non_exhaustive. That was what I originally intended but perhaps we got a bit sidetracked. The great thing about rust enums is that they are literally just discriminated unions. So size_of<Edition> will always be a single bytes, we could go up to edition 3227 and it will still only occupy one byte. Meanwhile storing the actual strings occupies the whole size of the string getting allocated + 32 bytes for storing the reference to it. Even a Cow doesn't help since it will occupy at least the size of the Owned type, so still 32 bytes + serde tends to prefer allocating but that's a completely different issue.

I still believe my original implementation on #174 is the best save for the fact that it only goes up to edition 2021, but that's an easy fix, we can add more variants and it would add no cost. Speaking  of which i believe similar optimization could be applied elsewhere (that's what I meant in #174). The current size of Metadata is around 200 bytes which might seem small but it adds up quickly, especially when you have to move around many instances of it, in a system with a basically non existent cpu cache.

> Making it an integer sounds fine to me :) thanks for the discussion, sorry it got a little acrimonious. I have a lot of sympathy for the embedded use case—hopefully switching to an integer will solve it.

I genuinely apologize about that, being a non native speaker it's kinda hard to read the tone of a conversation
(I actually had to google what acrimonious means xD)

oli-obk:
I wasn't offended fwiw.

Yea, just going with an enum that covers two or three future editions and bumping that in the future seems fine

PR #185, opened 2022-06-30
Made edition an enum again

morr0ne:
As mentioned in #183 this makes the Edition an enum again but also future proofs it by adding an `Unknown` variant

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -454,8 +454,9 @@ pub struct Target {
 
 /// The Rust edition
 ///
-/// As of writing this comment rust editions 2024, 2027 and 2030 are not actually a thing yet but are here nonetheless for future proofing.
+/// As of writing this comment rust editions 2024, 2027 and 2030 are not actually a thing yet but are parsed nonetheless for future proofing.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[non_exhaustive]
 pub enum Edition {
     /// Edition 2015
     #[serde(rename = "2015")]
@@ -466,15 +467,15 @@ pub enum Edition {
     /// Edition 2021
     #[serde(rename = "2021")]
     E2021,
-    /// Edition 2024
+    #[doc(hidden)]
     #[serde(rename = "2024")]
-    E2024,
-    /// Edition 2027
+    _E2024,
+    #[doc(hidden)]
     #[serde(rename = "2027")]
-    E2027,
-    /// Edition 2030
+    _E2027,
+    #[doc(hidden)]
     #[serde(rename = "2030")]
-    E2030,
+    _E2030,
 }
 
 impl Default for Edition {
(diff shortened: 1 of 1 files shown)


morr0ne:
This should be ready to merge but it would be a good idea to add some tests against all the enum variants. Right now only edition 2018 is tested. That ought to be enough but one can never be too sure. I'd love @oli-obk opinion on the tests as I'm a bit confused as how they are organized what the best way to test them would be.

sunshowers:
As I mentioned in #183, I'm afraid this has fundamental issues as explained in https://sunshowers.io/posts/open-closed-universes/#a-pattern-to-avoid-unknown-or-other-variants

morr0ne:
As highlighted by @sunshowers in #183 and their blog post this may have caused some issue when comparing existing values so I manually Implemented Eq/PartialEq/Hash so that cases like:
```rust
Edition::E2018 == Edition::Unknown("2018")
```
produce the correct value

jplatte:
I think it would be better if `Unknown` was `#[doc(hidden)] _Unknown(UnknownEdition)` where `UnknownEdition` is a struct with a private string field. Then you can add a `AsRef<str>` impl and maybe `.as_str()` method for people to use when they want to check for unknown editions.

morr0ne:
@jplatte I appreciate the input but most of the discussion has happened on #183, that might be a good idea but I think we might just end up discarding this whole pull request.

src/lib.rs
```diff
@@ -464,44 +466,17 @@ pub enum Edition {
...
-        match (self, other) {
-            (Self::Unknown(left), Self::Unknown(right)) => left == right,
-            (Self::E2015, Self::Unknown(unknown)) => unknown == "2015",
-            (Self::E2018, Self::Unknown(unknown)) => unknown == "2018",
-            (Self::E2021, Self::Unknown(unknown)) => unknown == "2021",
-            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
-        }
-    }
+    /// Edition 2024
+    #[serde(rename = "2024")]
```
oli-obk, line 470: Maybe hide the from the docs for now, it's not like matching on them explicitly helps anyone at present

morr0ne: I agree it probably doesn't make any sense to show them, I'll also mark it as non_exhaustive to make it less confusing

morr0ne: Maybe it's a good idea to also prefix them with an underscore for now?

oli-obk: hmmm... because we only need to parse them successfully, not make them accessible?

morr0ne: Yeah exactly, the edition 2024 and so forth don't exist yet so relying on that isn't right. We want to parse them so that old code will still parse correctly when a new edition comes out but it can't possibly make sense of it since it didn't exist at the time. When 2024 actually comes out we can remove the prefix unhide it from the docs

oli-obk:
Lgtm, i'll wait for @sunshowers to chime in before merging though

sunshowers:
Looks good to me, thanks for this!

PR #186, opened 2022-06-30
Added Changelog file

morr0ne:
Adds a simple changelog file to keep track of the various changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,35 @@
+# Changelog
+
+## Unreleased
+
+### Added
+
+- Re-exported `semver` crate directly.
+
+### Changed
+
+- Added `Unknown` variant to `Edition` enum to make it future proof.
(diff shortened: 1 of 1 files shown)


morr0ne:
Whoops. seems like a screwed up the branches and included #185 with the changes, Shouldn't be a problem if it gets merged first.

morr0ne:
The changelog makes use of a github integrated feature which should auto-magically link up the version with the appropriate tag/release. I never tested this before, but it should be as easy as publishing a release with the same version

morr0ne:
I fixed the accidental inclusion of #185 so now the only changes are the inclusion of the changelog file. It should be ready to merge without problems.

oli-obk:
Please rebase the commit away instead of reverting it

morr0ne:
That's the best I could do without access to my computer, I see know that was dumb. I'll fix it asap

Issue #193, opened 2022-07-02
Remove camino/Make it optional?


morr0ne:
Currently camino's Utf8PathBuf is used in the metadata object instead of the standard library PathBuf. However it seems rather strict to only allow utf8 paths since some more esoteric operating systems, such as windows, sometimes don't return valid utf8 data but valid paths nonetheless. I feel like utf8 validation should fall on the library consumer side and not enforced by this library. Maybe it could be made optional and enabled via a cargo feature?

jplatte:
Does cargo itself support non-utf8 paths? I was running under the assumption that it doesn't, and in that case I think using camino here is exactly the right thing to do.

morr0ne:
But what is the reason to use utf8 paths? I don't see why utf8 paths should be used in this context. They don't seem to provide any benefits and I had needless errors return because of it. Utf8 validation can still be performed after, but throwing away the whole metadata because some path wasn't utf8 doesn't make any sense.

jplatte:
One thing that I've found very convenient is that `Utf8Path` has a `Display` impl.

But aldo thinking more about it, almost all the camino paths used in this crate are deserialized from JSON, right? In that case, only utf-8 paths are possible anyways, as JSON is utf-8 and doesn't provide an escaped representation of non-utf8 bytes. serde_json will emit an error when you try to serialize a non-utf8 `Path`(`Buf`).

morr0ne:
> serde_json will emit an error when you try to serialize a non-utf8 `Path`(`Buf`).
If that's the case isn't camino even more redunant?

oli-obk:
There was previous discussion in https://github.com/oli-obk/cargo_metadata/pull/152

I don't know the best way forward. Some days I wish for implicit conversion from wrapper types to their wrapped types ^^ (not really, but still)



> If that's the case isn't camino even more redunant?

not quite. It's already possible to rely on the fact that the paths are utf8, what the use of camino types does here is to encode this information in the type system. I like this a lot, but I also see that nonstandard types are not always ideal. But if no one adopts new helper crates and types, they can't really become standard-by-convention. So... 🤷 good arguments for all sides, but the main forker (rust-analyzer) will fork either way, so there isn't really any loss here.

morr0ne:
[...]

PR #194, opened 2022-07-02
Made camino an optional dependency.

morr0ne:
As I already said in #193 I don't feel like camino is necessary however there might people that relay/need this kind of behavior so I made it an optional dependency disabled by default. I also updated the changelog with the corresponding entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - Made `parse_stream` more versatile by accepting anything that implements `Read`.
+- Made `camino` an optional dependency disabled by default.
 
 ### Removed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 rust-version = "1.42.0"
 
 [dependencies]
-camino = { version = "1.0.7", features = ["serde1"] }
+camino = { version = "1.0.7", optional = true, features = ["serde1"] }
 cargo-platform = "0.1.2"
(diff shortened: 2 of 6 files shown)


taiki-e:
This does not seem right because [the cargo feature should be additive](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

For example, enabling this feature may break dependencies that do not enable this feature because some signatures of camino and std::path APIs are incompatible. Also, in some cases, third-party traits implemented in std::path are not implemented in camino, so the code that uses them will also be broken.

morr0ne:
I assumed Utf8PathBuf implemented all the methods from PathBuf but I see how that can be a problem especially with third-party impls. Is there a better solution to make camino optional? Otherwise it might be better to remove camino entirely. As I said in #193 it hads needless restrictions considering that every Utf8PathBuf is a valid PathBuf but not the other way around.

oli-obk:
> I assumed Utf8PathBuf implemented all the methods from PathBuf but I see how that can be a problem especially with third-party impls.

That's not the main issue, the main issue is that the types are different, so if someone passes that field to a function expecting a `PathBuf`, because their crate uses `cargo_metadata` without the camino feature, then that would break the moment any other crate in the deptree enables the feature.

Will comment on the issue

morr0ne:
Yeah I see how that's a big very big issue. Maybe a better solution is a wrapper type kinda like [the map in serde_jsom](https://docs.rs/serde_json/latest/serde_json/struct.Map.html). But honestly I think camino should be removed all together.

PR #199, opened 2022-08-01
Add `Edition::as_str()` method

djc:
In skeptic, I'd like to be able to pass an `Edition` in a command-line flag to rustc.

Could potentially also add a `fmt::Display` impl.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -477,6 +477,21 @@ pub enum Edition {
     _E2030,
 }
 
+impl Edition {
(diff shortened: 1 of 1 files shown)


oli-obk:
Well... you can already serialize to `String` via serde 😆 

I think a `Display` impl is most useful, yes.

djc:
I added a `Display` impl. I'd prefer to also keep the `as_str()` API because it enables getting the string version without allocating.

PR #202, opened 2022-08-21
Fix `Package::readme()` to return correct path

### LINKED ISSUE #201: `Package::readme()` is broken
`Package::readme()` returns invalid path like `/path/to/package/Cargo.toml/README.md`.

https://github.com/oli-obk/cargo_metadata/blob/ea73f56dd64c381574c5b670022d602e43205af5/src/lib.rs#L365-L382

The `readme` path from the TOML file must be combined with `manifest_path.parent().unwrap()`, not `manifest_path`.

gifnksm:
fixes #201

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -375,9 +375,12 @@ impl Package {
 
     /// Full path to the readme file if one is present in the manifest
     pub fn readme(&self) -> Option<Utf8PathBuf> {
-        self.readme
-            .as_ref()
-            .map(|file| self.manifest_path.join(file))
+        self.readme.as_ref().map(|file| {
+            self.manifest_path
+                .parent()
+                .unwrap_or(&self.manifest_path)
+                .join(file)
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -375,9 +375,12 @@ impl Package {
 
     /// Full path to the readme file if one is present in the manifest
     pub fn readme(&self) -> Option<Utf8PathBuf> {
-        self.readme
-            .as_ref()
-            .map(|file| self.manifest_path.join(file))
+        self.readme.as_ref().map(|file| {
+            self.manifest_path
+                .parent()
+                .unwrap_or(&self.manifest_path)
```
jyn514, line 381: why this `unwrap_or`? shouldn't we return `None` if the manifest path is `/`? not sure how that would happen but still ...

jyn514 (approved):

src/lib.rs
gifnksm: I'm borrowing the implementation from `Package::license()`, so I don't know why unwrap_or is used...

https://github.com/oli-obk/cargo_metadata/blob/ea73f56dd64c381574c5b670022d602e43205af5/src/lib.rs#L367-L374

I will modify `Package::license()` and `Package::readme()` to use `unwrap` instead of `unwrap_or` if you like.

jyn514: > I'm borrowing the implementation from `Package::license()`, so I don't know why `unwrap_or` is used...

ah ok, it's fine to keep it as-is then :) thanks for clarifying!

jyn514 (approved):

PR #204, opened 2022-09-09
impl `Copy` for `Edition`

msrd0:


jplatte (commented):
I don't think this is necessarily a good idea. It makes extra promises that aren't really that important, and would have to be removed again if `Edition` was changed to allow arbitrary edition strings (which would not necessarily be a breaking change).

msrd0:
If we can make the assumption that editions will continue to be named after the year they are introduced in, this shouldn't be a problem.

oli-obk:
I'm fine making it `Copy`, if we switch to a more stringly typed representation we are likely at a stage where a new major bump is a good idea anyway.

PR #206, opened 2022-09-29
reexport a few types from serde_json

pacak:
Mostly to be able to consume Metadata::workspace_metadata without having to worry about depending on serde_json directly and keeping dependencies in sync.

Well, you can use methods on `Value` as is, but this gives you pattern matching.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,6 +92,7 @@ use std::str::from_utf8;
 pub use camino;
 pub use semver;
 use semver::{Version, VersionReq};
+pub use serde_json::{Map, Number, Value};
 
 pub use dependency::{Dependency, DependencyKind};
 use diagnostic::Diagnostic;


oli-obk:
`serde_json` never does major or minor version bumps, only patch version bumps, so all `serde_json` versions are compatible.

pacak:
> `serde_json` never does major or minor version bumps

https://crates.io/crates/serde_json/versions - 0.5, 0.6, 0.7, etc.

It might be true for now but _never_ is a very long time. People use rust so compiler doesn't let them to make mistakes rather than relying on good old "I never write use after free in C++".

oli-obk:
Those are versions that don't even work with `serde` 1.0 . If there will be a breaking serde_json bump, the impact on the ecosystem will be immense and we'll still be impacted. I don't think there's any benefit to doing this *specifically for serde*. For many other crates I would agree, but the stability guarantees of serde are quite extreme.

> People use rust so compiler doesn't let them to make mistakes rather than relying on good old "I never write use after free in C++".

This seems rather unrelated. The only problem here is if you update a dependency version but not the other, there is no soundness or correctness concern, you'd just get a compilation error.

pacak:
:shrug:

PR #211, opened 2022-11-04
Allow `MetadataCommand` to inherit stderr from parent

### LINKED ISSUE #210: Allow showing output
xref https://github.com/PyO3/maturin/issues/1243

Commands using `cargo_metadata` can't show the cargo output, which makes problems pretty hard to diagnose if something goes wrong, like cargo being blocked.

For example, when building a Python extension project with maturin using `maturin build`, and a lock file `~/.cargo/.package-cache` exists, the command will show no output, and stay idle forever. Being able to show the cargo output `Blocking waiting for file lock on the registry index` to the user would be extremely helpful.

messense:
Closes #210

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,7 +86,7 @@ use std::env;
 use std::fmt;
 use std::hash::Hash;
 use std::path::PathBuf;
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::str::from_utf8;
 
 pub use camino;
@@ -592,6 +592,8 @@ pub struct MetadataCommand {
     /// Arbitrary command line flags to pass to `cargo`.  These will be added
     /// to the end of the command line invocation.
     other_options: Vec<String>,
+    /// Show stderr
+    verbose: bool,
 }
 
 impl MetadataCommand {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -687,6 +689,12 @@ impl MetadataCommand {
         self
     }
 
+    /// Set whether to inherit stderr from parent
+    pub fn inherit_stderr(&mut self, inherit: bool) -> &mut MetadataCommand {
```
oli-obk, line 693: That's what happens underneath, but I think the API should be more about how it appears for users. So something about printing stderr? or showing errors?

Do we even need this? We could alternatively return stderr from `exec` so it can be accessed as a string

messense: As the issue reported, cargo could be blocked for a long time, capture stderr then print does not work because it will be only be printed after the file lock is released.

oli-obk: ah right, that was the use case, forgot again. thanks.

So only my first point of naming this better stands.

messense: Originally I used `show_output` but it doesn't say what it will output. Do you have any suggestions in mind?

oli-obk: Maybe just `verbose`? Idk, I don't have a good idea either. Even `show_stderr` seems better than `inherit_stderr` tho

messense: `verbose` sounds good.

PR #213, opened 2022-11-20
Add `#[serde(default)]` to `Artifact::manifest_path`

### LINKED ISSUE #212: Failed to parse compiler artifact output on Rust 1.48.0
```
{
  "reason": "compiler-artifact",
  "package_id": "pyo3-ffi 0.17.3 (path+file:///home/runner/work/pyo3/pyo3/pyo3-ffi)",
  "target": {
    "kind": [
      "lib"
    ],
    "crate_types": [
      "lib"
    ],
    "name": "pyo3-ffi",
    "src_path": "/home/runner/work/pyo3/pyo3/pyo3-ffi/src/lib.rs",
    "edition": "2018",
    "doctest": true,
    "test": true
  },
  "profile": {
    "opt_level": "0",
    "debuginfo": 2,
    "debug_assertions": true,
    "overflow_checks": true,
    "test": false
  },
  "features": [
    "default",
    "extension-module"
  ],
  "filenames": [
    "/home/runner/work/pyo3/pyo3/target/x86_64-unknown-linux-gnu/debug/deps/libpyo3_ffi-20b19129a10dabc9.rlib",
    "/home/runner/work/pyo3/pyo3/target/x86_64-unknown-linux-gnu/debug/deps/libpyo3_ffi-20b19129a10dabc9.rmeta"
  ],
  "executable": null,
  "fresh": false
}
```

The above json doesn't contain the `manifest-path` field, after https://github.com/oli-obk/cargo_metadata/pull/195 it can't be parsed into a `Message::Artifact` and becomes `Message::TextLine`.

messense:
Fixes #212

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -35,6 +35,7 @@ pub struct Artifact {
     /// The package this artifact belongs to
     pub package_id: PackageId,
     /// Path to the `Cargo.toml` file
+    #[serde(default)]
     pub manifest_path: Utf8PathBuf,
     /// The target this artifact was compiled for
     pub target: Target,


oli-obk:
This is a breaking change. If we're going to do one anyway we could also just bump msrv to 1.50 which seems to be what we support in practice.

What we could land is adding a `#[serde(default)]` attribute to the field as that is backwards compatible

messense:
> What we could land is adding a `#[serde(default)]` attribute to the field as that is backwards compatible

Done.

PR #214, opened 2023-01-12
Release via GitHub Actions, with semver-checking and git tag.

obi1kenobi:
Add a [manually-triggered workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) for releasing a new version of `cargo_metadata`, complete with build caching, tagging, and cargo-semver-checks.

This uses a "bare" tag format like `0.15.2`, following the example of the most recent tags in the project. Some of the older tags use a `v`-prefixed tag name like `v0.13.0`; if that's a preferable format then I'd be happy to switch it.

It's also possible to make a workflow that is automatically triggered once `Cargo.toml` is updated with a new version on the `main` branch. This is how I generally configure my projects (e.g. [that's how `cargo-semver-checks` releases](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/workflows/release.yml)), but I wasn't sure if that would also be your preference so I went with a manual trigger instead.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,76 @@
+name: Release new version
+
+on:
+  workflow_dispatch:
+    secrets:
+      CARGO_REGISTRY_TOKEN:
+        required: true
+
+env:
+  RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
+
+jobs:
+  create-release:
+    name: Create release
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: true
+
+      - name: Install rust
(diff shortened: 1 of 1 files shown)


obi1kenobi:
I tested all the bash components of this manually on my machine, and as best I can tell they appear to work.

GitHub Actions to my knowledge doesn't have great facilities for testing workflows, so the best way for a real-world test is to publish a new release. Whenever one is ready, please feel free to either tag me here or on [Mastodon](https://hachyderm.io/@predrag) or [Twitter](https://twitter.com/PredragGruevski) and I'd be happy to monitor the publish process and fix anything that might be broken.

Thanks again for this awesome crate and for all the awesome work you're doing in the Rust ecosystem!

oli-obk:
I like doing it manually for now, but seems reasonable to do it just by looking at the Cargo.toml.

PR #216, opened 2023-01-30
Allow setting env vars in cargo_metadata command

illicitonion:
Just like with `other_options`, this is a reasonably common use-case, and needing to re-implement `exec` just to pass an extra env var is a bit annoying.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -710,7 +710,11 @@ impl MetadataCommand {
     ///     // ...
     ///     # ;
     /// ```
-    pub fn env<K: Into<OsString>, V: Into<OsString>>(&mut self, key: K, val: V) -> &mut MetadataCommand {
+    pub fn env<K: Into<OsString>, V: Into<OsString>>(
+        &mut self,
+        key: K,
+        val: V,
+    ) -> &mut MetadataCommand {
         self.env.insert(key.into(), val.into());
         self
     }


oli-obk:
Do you have any general use cases that we could document as an example?

illicitonion:
Absolutely - the two big ones I had were:
1. Setting cargo config values - as an example I wanted to conditionally set `CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` to test out sparse indexes, based on some other config in my program or the version of Cargo being used, or setting `CARGO_NET_GIT_FETCH_WITH_CLI` based on some other logic.
2. In alternate build systems, specifying a custom value of `RUSTC` because it may not be where Cargo expects.

Where do you think would be the best place to document this? In doc comments (either as prose or code?), or elsewhere?

oli-obk:
In doc comments on the env method seems best to me. Both of these examples are great and different enough for it to make sense to include both.

illicitonion:
Added some example prose and an example code snippet

src/lib.rs
```diff
@@ -694,6 +694,21 @@ impl MetadataCommand {
...
+    /// 1. Setting cargo config values without needing a .cargo/config.toml file, e.g. to set
+    ///    `CARGO_NET_GIT_FETCH_WITH_CLI=true`
+    /// 2. To specify a custom path to RUSTC if your rust toolchain components aren't laid out in
+    ///    the way cargo expects by default.
+    ///
+    /// ```no_run
+    /// # use cargo_metadata::{CargoOpt, MetadataCommand};
+    /// MetadataCommand::new()
+    ///     .env(String::from("CARGO_NET_GIT_FETCH_WITH_CLI"), String::from("true"))
+    ///     .env(String::from("RUSTC"), String::from("/path/to/rustc"))
```
oli-obk, line 708: This example makes me think we should make `env` accept `impl Into<String>` or even just `impl Display` (the extra allocation in case of passing a `String` for the latter seems ok for this crate)

illicitonion: Switched to use `Into<OsString>` because that's closer to what `std::process::Command` uses, but happy to switch to `Into<String>` if you'd prefer.

illicitonion: Switched to use `Into<OsString>` because that's closer to what `std::process::Command` uses, but happy to switch to `Into<String>` if you'd prefer.

oli-obk:
Rustfmt is unhappy, but otherwise this lgtm

illicitonion:
Fixed - thanks!

PR #221, opened 2023-03-03
Use `v` prefix for version tags.

obi1kenobi:
I just noticed that the vast majority of version tags in this repo use the `v` prefix for version tags, but this workflow doesn't add that prefix right now.

Perhaps it's better to stay consistent as much as possible.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
         run: |
           # Fail on first error, on undefined variables, and on errors in pipes.
           set -euo pipefail
-          git tag "${{ steps.version-check.outputs.version }}"
+          git tag "v${{ steps.version-check.outputs.version }}"
           git push origin "${{ steps.version-check.outputs.version }}"
 
       - uses: taiki-e/create-gh-release-action@v1


oli-obk:
I only did that because of muscle memory from another crate. Now that we've done releases without it let's stick with it?

obi1kenobi:
> I only did that because of muscle memory from another crate. Now that we've done releases without it let's stick with it?

Sounds good.

If you'd like, I can also push prefix-less tags for all the current tags that have a `v` prefix, without deleting the old ones. Not sure if the duplicated-but-consistent tags for each version would be better than non-duplicated but mixed-prefix ones.

oli-obk:
> Not sure if the duplicated-but-consistent tags for each version would be better than non-duplicated but mixed-prefix ones.

seems fine to have the duplication, but we could also delete the prefixed tags if github allows that

obi1kenobi:
It does allow deleting tags and I'm happy to do it if you think it's the way to go.

The only concern I have is that it might break any workloads that were relying on the tag existing. This is a much bigger concern for GitHub Actions than for a library crate -- I'd be surprised if anyone is installing from a git tag -- but I wanted to bring it up in case there's a edge case I've missed.

oli-obk:
oh yea, I am not worried, if folks want to rely on that they can grab the crates.io commit hash

PR #222, opened 2023-03-30
Explicitly depend on serde_derive

Veykril:
This allows `serde` and `serde_derive` to build in parallel, improving compile times (once the same change happens in `camino` which this crate depends on), see https://github.com/rust-lang/rust-analyzer/pull/14450


Veykril:
That is a shame but understandable 😕

PR #223, opened 2023-03-31
Make serialization of packages deterministic.

matts1:
The nondeterminism is causing cache misses for users of bazel, who depend on this for rules_rust's crate_universe package (https://github.com/bazelbuild/rules_rust/tree/main/crate_universe).


src/lib.rs
```diff
@@ -311,6 +323,7 @@ pub struct Package {
     /// Targets provided by the crate (lib, bin, example, test, ...)
     pub targets: Vec<Target>,
     /// Features provided by the crate, mapped to the features required by that feature.
+    #[serde(serialize_with = "sorted_map")]
     pub features: HashMap<String, Vec<String>>,
```
jplatte, line 327: Personally I would prefer if the type of this field was simply changes, but **I'm not a maintainer**.

oli-obk: I agree. We can do this when we do the next major version bump. I'll open an issue for it

PR #227, opened 2023-04-24
Use the new cargo-semver-checks GitHub Action

obi1kenobi:
It does a lot of stuff "correctly" given that it knows it's running on GitHub. For example:
- downloads a prebuilt binary, which is much faster than `cargo install`
- caches previous versions' rustdoc JSONs to speed up builds

More details here:
https://github.com/obi1kenobi/cargo-semver-checks-action#cargo-semver-checks-action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,14 +46,10 @@ jobs:
             exit 1
           fi
 
-      # TODO: Replace this with the cargo-semver-checks v2 GitHub Action when it's stabilized:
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Also, the new cargo-semver-checks v0.20 is more than 2000x faster on large crates thanks to the new optimization API in Trustfall v0.4: https://twitter.com/PredragGruevski/status/1650493414197059587

oli-obk:
Cool thanks! will migrate this out to my other crates using the script ^^

PR #228, opened 2023-05-01
feat: use BTreeMap for deterministic serialization

### LINKED ISSUE #224: Replace all HashMaps with BTreeMaps for determinism
Personally I would prefer if the type of this field was simply changes, but **I'm not a maintainer**.

_Originally posted by @jplatte in https://github.com/oli-obk/cargo_metadata/pull/223#discussion_r1154095243_

SteveLauC:
## What does this PR do
1. Replace `HashMap` with `BTreeMap` in `struct Package` for deterministic serialization
2. `HashMap`  used in `struct MetaCommand` is also replaced with `BTreeMap`
    > This won't affect serialization as this type is not `#[derive(Serialize)]`
3. Remove two extra `std::` prefix in `std::fmt` as it is already imported

## Issue
Closes #224

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,7 @@
 use camino::Utf8PathBuf;
 #[cfg(feature = "builder")]
 use derive_builder::Builder;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::env;
 use std::ffi::OsString;
 use std::fmt;
@@ -109,7 +109,7 @@ pub use messages::{
     ArtifactBuilder, ArtifactProfileBuilder, BuildFinishedBuilder, BuildScriptBuilder,
     CompilerMessageBuilder,
 };
-use serde::{Deserialize, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 
 mod dependency;
 pub mod diagnostic;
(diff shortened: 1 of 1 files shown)


oli-obk:
I think we should use this opportunity and add clippy to our CI and then deny hash maps via https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_types

Would you like to tackle this as well? If not, that's ok, we can land the PR as is.

SteveLauC:
> Would you like to tackle this as well? If not, that's ok, we can land the PR as is.

Happy to include that in this PR, will do it later (tonight or tomorrow) :)

clippy.toml
```diff
@@ -0,0 +1,3 @@
+disallowed-types = [
+    { path = "std::collections::HashMap", reason = "HashMap's key order is unspecified and thus serializing such a type will result in a random output, use BTreeMap instead." }
```
SteveLauC, line 2: I put that long `reason` in a single line as newline is [still not allowed in `inline table`](https://github.com/toml-lang/toml/blob/913ff560d935e7fb2788b402abba567680f6e03c/CHANGELOG.md#unreleased)

If we choose to use multiple line string, it would look like this:

```tom
disallowed-types = [
    { path = "std::collections::HashMap", reason = """
      HashMap's key order is unspecified and thus serializing such a type will
      result in a random output, use BTreeMap instead.
      """ }
]
```
which is also not good:(

PR #230, opened 2023-05-25
bump derive_builder dependency to v0.12

decathorpe:
Version 0.12.0 of the `derive_builder` crate introduces only a small breaking change which does not affect this crate.

I have also bumped the MSRV of `cargo_metadata` to 1.56.0 - I tried compiling this crate with the currently set 1.42.0, which fails (even before bumping derive_builder to v0.12) because some of this crate's dependencies have moved to the Rust 2021 edition (which was stabilized in Rust 1.56.0).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ description = "structured access to the output of `cargo metadata`"
 license = "MIT"
 readme = "README.md"
 edition = "2018"
-rust-version = "1.42.0"
+rust-version = "1.56.0"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


oli-obk:
sgtm. You need to address the clippy failures though :)

Maybe we should also move this crate to the 2021 edition

decathorpe:
The clippy warning should now be addressed. I'm not sure if bumping to the 2021 edition would be necessary, unless you plan to use some features that are 2021-or-later-only.  But at least it won't necessitate bumping the MSVR.

PR #235, opened 2023-05-30
Allow bool for readme in manifest

### LINKED ISSUE #234: readme can also be a boolean
According to [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field), you can specify `readme = false` to disable readme lookups. I had opened https://github.com/rustwasm/wasm-pack/pull/1216 to fix this, but they took a dependency on this crate that also declares it as only an optional string.

https://github.com/oli-obk/cargo_metadata/blob/010cd78751e51dc48d0ca08ca0f4be3ac054a689/src/lib.rs#L324

heaths:
Fixes #234

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -321,7 +321,7 @@ pub struct Package {
     #[serde(default)]
     pub keywords: Vec<String>,
     /// Readme as given in the `Cargo.toml`
-    pub readme: Option<Utf8PathBuf>,
+    pub readme: Option<Readme>,
     /// Repository as given in the `Cargo.toml`
     // can't use `url::Url` because that requires a more recent stable compiler
     pub repository: Option<String>,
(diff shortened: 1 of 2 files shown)


ehuss:
I'm a bit confused about this change. Cargo doesn't expose the true/false status as a boolean via metadata. It should only ever be a string or null. Can you say more about what exact scenario this is trying to address?

ehuss:
Sorry, maybe I wasn't clear. The true/false value of the readme field isn't exposed by cargo (see [`cargo metadata`](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html)). This crate is just representing the values that cargo itself exposes via the `cargo metadata` command. A value of `readme=false` is translated to null, and a value of `true` is translated to "README.md".

heaths:
Ah. Thanks for clarifying. I looked through the project de/serialization a bit but didn't realize you were effectively parsing the output of a command. Given the "translation" you mentioned, that should still be enough to go on to determine if the readme resolution was disabled, though.

Thank you.

jplatte:
Well, if the `cargo metadata` output indeed makes a difference here (`null` vs. an absent field), I think it would be reasonable for this crate to reflect that. I wrote the `js_option` crate for exactly this use case of distinguishing null vs. missing, though it might not be desireable to take on the extra public dependency here.

ehuss:
`cargo metadata` does not distinguish a difference (the field is always present).

PR #236, opened 2023-06-14
Replace `thiserror` with manual impl

bheylin:
Using `thiserror` to define a single `Error` is overkill and forces your users to wait on `syn` to compile.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,4 +1,5 @@
-use std::{io, str::Utf8Error, string::FromUtf8Error};
+use core::fmt;
+use std::{error::Error as StdError, io, str::Utf8Error, string::FromUtf8Error};
 
 /// Custom result type for `cargo_metadata::Error`
 pub type Result<T> = ::std::result::Result<T, Error>;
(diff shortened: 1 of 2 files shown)


src/errors.rs
```diff
@@ -21,32 +22,75 @@ pub type Result<T> = ::std::result::Result<T, Error>;
...
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::CargoMetadata { stderr } => write!(f, "`cargo metadata` exited with an error: {stderr}"),
+            Error::Io(err) => write!(f, "failed to start `cargo metadata`: {err}"),
+            Error::Utf8(err) => write!(f, "cannot convert the stdout of `cargo metadata`: {err}"),
+            Error::ErrUtf8(err) => write!(f, "cannot convert the stderr of `cargo metadata`: {err}"),
+            Error::Json(err) => write!(f, "failed to interpret `cargo metadata`'s json: {err}"),
+            Error::NoJson => f.write_str("could not find any json in the output of `cargo metadata`"),
+}
```
jplatte, line 94: Indentation fail?

taiki-e:
> wait on `syn` to compile

Note that this PR does not solve this problem because this crate depends on syn even via serde's derive feature.

bheylin:
> > wait on `syn` to compile
> 
> Note that this PR does not solve this problem because this crate depends on syn even via serde's derive 

I see it as a step towards solving the problem

src/errors.rs
bheylin: Hmm usually RA fixes these issues up for me.
Anyways, it's resolved in https://github.com/oli-obk/cargo_metadata/pull/236/commits/88f9b71c28b18ce48c4cca51f52d31a241d60c92

jplatte:
> I would ultimately like to make serde and serde-json optional deps.

How? This crate's purpose is to parse the JSON output from `cargo metadata`, do you want to use a different library for the parsing? That seems like it should be discussed in an issue first.

bheylin:
> > I would ultimately like to make serde and serde-json optional deps.
> 
> How? This crate's purpose is to parse the JSON output from `cargo metadata`, do you want to use a different library for the parsing? That seems like it should be discussed in an issue first.

For sure, the serde issue is a bigger issue, and potential one that's not worth it.
We will for sure discuss that in an issue.
Let's step back to my larger problem.

I have various projects using an `xtask` style binary for building and various other tasks.
I'm trying to keep the deps to a minimum for this case.

So for me, removing `thiserror` at least lightens the dep load.
Yes on it's own it doesn't removed the link with `syn`.
But I still see using `thiserror` for defining a single `Error` as overkill.

jplatte:
Right. I wouldn't disagree entirely, but I think the maintenance effort / user value tradeoff isn't as clear when syn is required anyways. Let's see what the actual maintainers think 😄

PR #238, opened 2023-06-26
doc: state inner repr of `Source` is opaque

### LINKED ISSUE #237: Why is `PackageId` opaque but `SourceId` not?
In https://github.com/rust-lang/cargo/pull/12313, the Cargo team is discussing `PackageId` and `SourceId` should be seen as being opaque. We found that for `PackageId` in `cargo_metadata` crate [it is mentioned as opaque](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.PackageId.html), but for `SourceId` [it's not](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Source.html). I wonder if there was a reason this crate didn't say the inner representations of `SourceId` is also opaque, or you've spotted some valid use cases. Thank you :)

weihanglo:
Fixes #237

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,6 +117,7 @@ mod errors;
 mod messages;
 
 /// An "opaque" identifier for a package.
+///
 /// It is possible to inspect the `repr` field, if the need arises, but its
 /// precise format is an implementation detail and is subject to change.
 ///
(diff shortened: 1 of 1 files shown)


oli-obk:
Thanks!

PR #241, opened 2023-07-19
Update debuginfo for change in 1.71

### LINKED ISSUE #240: Non numerical values for debuginfo cannot be parsed from cargo messages
Valid values for debuginfo level now also include strings "line-tables-only" and "line-directives-only", see [here](https://doc.rust-lang.org/cargo/reference/profiles.html#debug). 

`cargo` will emit those as part of json output. `profile` for an can look artifact look like this

```json
"profile":{"opt_level":"3","debuginfo":"line-tables-only","debug_assertions":false,"overflow_checks":false,"test":false},
```
I'm trying to parse it into `cargo_metadata::messages::Message` datatype and it fails to parse it, probably because 
[`ArtifactProfile`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.ArtifactProfile.html) defines `debuginfo` as `Option<u32>`

https://github.com/pacak/cargo-show-asm/issues/187 - related ticket

ehuss:
There was an unanticipated change in 1.71 that added stringly-typed debuginfo in the profile struct. After discussing with the cargo team, we have decided to keep this change even though it has caused breakage. Fortunately this only affects projects which have opted-in to the string-based variants. 

There isn't a clear alternative for how to represent this information (they can't be accurately mapped to numbers), and we don't have a way to version this JSON output (which would also be a breaking change if we changed the version). We currently don't have a good story for how to manage JSON compatibility of the compiler messages, which is something we should probably figure out at some time.

I decided to go ahead and map the values to their semantic meaning, with an escape hatch for unknown values that might be added in the future. Alternatively it could just be a two-variant enum (int/string), but adding the mapping wasn't too difficult. However, this means that the round-trip serialization isn't exact (null or missing gets mapped to 0). I don't expect that to be an issue, but I'm not sure why anyone would serialize these.

Fixes #240

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,7 +102,8 @@ pub use errors::{Error, Result};
 #[allow(deprecated)]
 pub use messages::parse_messages;
 pub use messages::{
-    Artifact, ArtifactProfile, BuildFinished, BuildScript, CompilerMessage, Message, MessageIter,
+    Artifact, ArtifactDebuginfo, ArtifactProfile, BuildFinished, BuildScript, CompilerMessage,
+    Message, MessageIter,
 };
 #[cfg(feature = "builder")]
 pub use messages::{
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -2,8 +2,8 @@ use super::{Diagnostic, PackageId, Target};
 use camino::Utf8PathBuf;
 #[cfg(feature = "builder")]
 use derive_builder::Builder;
-use serde::{Deserialize, Serialize};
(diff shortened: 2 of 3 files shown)


obi1kenobi (approved):
I've been following the debuginfo zulip thread, and this solution looks great to me :+1:

oxalica:
This affects many downstream crates. I just ran into an suspicious error with [cargo-flamegraph](https://github.com/flamegraph-rs/flamegraph) and got here finally. Hope to have a new release soon.

PR #243, opened 2023-07-20
fix!: change type of Package.rust_version to BareVersion

not-my-profile:
As per the Cargo Book[1]:

> The Rust version must be a bare version number with two or three
> components; it cannot include semver operators or pre-release
> identifiers.

VersionReq therefore obviously is the wrong type since it describes[2]:

> the intersection of some version comparators such as `>=1.2.3, <1.8`

Since the semver crate apparently doesn't include a type for bare versions, this commit introduces a new such type. Prior art includes the cargo-msrv crate which already defines its own BareVersion type.[3] The type introduced by this commit differs in that it's a struct instead of an enum since that allows easier access to the major and minor numbers.

[1]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
[2]: https://docs.rs/semver/1.0.7/semver/struct.VersionReq.html
[3]: https://github.com/foresterre/cargo-msrv/blob/v0.15.1/src/manifest/bare_version.rs

### DIFF
--- a/src/bare_version.rs
+++ b/src/bare_version.rs
@@ -1,0 +1,159 @@
+//! This module defines the [`BareVersion`] type used for `rust_version` in [`Package`](crate::Package).
+use std::{fmt::Display, num::ParseIntError, str::FromStr};
+
+use semver::Version;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// A bare version number with two or three components (used for `rust_version` in [`Package`](crate::Package)).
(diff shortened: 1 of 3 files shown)


oli-obk:
I'd rather not have this type live in this crate. Maybe this is a good type for addition in `semver`?

weihanglo (commented):
Perhaps `semver::Version` is a good fit for this purpose.

PR #244, opened 2023-07-20
Change type of Package.rust_version from VersionReq to Version

not-my-profile:
As per the Cargo Book[1] the `rust-version` must:

> be a bare version number with two or three components;
> it cannot include semver operators or pre-release identifiers.

VersionReq therefore obviously is the wrong type since it describes[2]:

> the intersection of some version comparators such as `>=1.2.3, <1.8`

It doesn't make much sense to deserialize strings that explicitly "cannot include semver operators" as a vector of comparison operators.

While the Version type's Deserialize implementation requires three components, we can just use a custom deserialize function to deserialize 1.2 as 1.2.0. This commit introduces such.

[1]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
[2]: https://docs.rs/semver/1.0.7/semver/struct.VersionReq.html


not-my-profile:
Yes it does. I tested it ... `cargo metadata` does not normalize such truncated version numbers in its output.

PR #248, opened 2023-08-23
Add `workspace_default_members` field

### LINKED ISSUE #215: `workspace.default-members` key is missing from output
According to the [doc of `cargo build`](https://doc.rust-lang.org/cargo/commands/cargo-build.html):

> The default members of a workspace can be set explicitly with the `workspace.default-members` key in the root manifest.

However there's no such field in the output of `cargo metadata`.

Is there a workaround to this other than parsing the toml file?

sorcio:
Fixes #215

* Adding new field `workspace_default_members` which was added in Cargo 1.71. I made it an option because we have no way to reconstruct the value in previous Cargo versions.
* I also added a convenience method to collect `&Package`s that are default workspace members. But it also needs to either return an Option or panic. I'm not sure this is ideal, because at some point in the future the minimum supported Cargo version might be >=1.71 and changing this API would be a backwards incompatible change. But panicking doesn't feel ideal either. Should I remove this method?
* I did not want to bump the minimum version for the `all_the_fields` test. Instead I added a conditional assertion, following the same pattern introduced by the `rust_version` field.
* There are no tests that exercise this field other than "it should be the same as workspace_members". Should I add some test data?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,10 +151,11 @@ pub struct Metadata {
     pub packages: Vec<Package>,
     /// A list of all workspace members
     pub workspace_members: Vec<PackageId>,
-    /// A list of all workspace members
+    /// The list of default workspace members
     ///
-    /// This is always `None` if running with a version of Cargo older than 1.71.
-    pub workspace_default_members: Option<Vec<PackageId>>,
+    /// This not available if running with a version of Cargo older than 1.71.
+    #[serde(skip_serializing_if = "workspace_default_members_is_missing")]
+    pub workspace_default_members: WorkspaceDefaultMembers,
     /// Dependencies graph
     pub resolve: Option<Resolve>,
     /// Workspace root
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -151,6 +151,10 @@ pub struct Metadata {
     pub packages: Vec<Package>,
     /// A list of all workspace members
     pub workspace_members: Vec<PackageId>,
+    /// A list of all workspace members
+    ///
+    /// This is always `None` if running with a version of Cargo older than 1.71.
+    pub workspace_default_members: Option<Vec<PackageId>>,
```
oli-obk, line 157: Is this `Vec` ever empty? If not, then we could avoid the `Option` and use an empty vec as the value signifying that cargo is too old

oli-obk: ok, in that case create a `WorkspaceDefaultMembers` struct that has a private field storing the `Option<Vec<PackageId>>`.

We then have a few options:

* implement `Deref` for it, panicking if it is none, otherwise returning `&[PackageId]`
* keep your `workspace_default_packages` method
* use the same method, but don't return an `Option`, by populating it with the right values (just grab the `workspace_members` list?)
* do some shenanigans during deserialization of `Metadata`, copying `workspace_members` if `workspace_default_members` does not exist in the serialized data.

opinions?

oli-obk: Oh. I wasn't aware of this issue. Thanks.

I guess let's go with one of the panicking versions then?

oli-obk:
> would it be interesting to expose something similar in spirit to `cargo_version()` as public API?

I don't know where we'd get this information from and how we'd use it. Do you have ideas?

PR #252, opened 2023-10-09
support libtest json output

### LINKED ISSUE #251: `cargo test -- --format json` support?
Wondering if you'd be willing to expand it, to support `cargo test`, specifically `cargo test --message-format json -- -Zunstable-options --report-time --show-output --format json` output.

Ive implemented the test part already, in my [cargo-kewl](https://github.com/bend-n/cargo-kewl/blob/master/src/cargo.rs#L53-L80) project, so id be willing to make a PR for it.

bend-n:
closes #251


src/libtest.rs
```diff
@@ -0,0 +1,173 @@
...
+        match self {
+            Self::Ok { stdout, .. } | Self::Failed { stdout, .. } => stdout.as_deref(),
+            _ => None,
+        }
+    }
+
+    /// If this test is in progress.
+    pub fn in_progress(&self) -> bool {
+        matches!(self, Self::Started { .. })
+    }
```
oli-obk, line 113: I don't think this is too useful, as something like `in_progress` sounds like it could apply to intermediate messages, too, but those don't exist.

src/lib.rs
```diff
@@ -115,6 +116,7 @@ use serde::{Deserialize, Deserializer, Serialize};
 mod dependency;
 pub mod diagnostic;
 mod errors;
+pub mod libtest;
```
oli-obk, line 119: Since this may change in rustc with any rustc release, what do you think about adding an `unstable` cargo feature to `cargo_metadata` and hiding this module behind it?

bend-n: Seems like a good idea

src/libtest.rs
bend-n: Yeah, I never actually needed this.

oli-obk:
Thanks! Lgtm. If you want I can publish a minor version bump immediately via github actions, but you'll have to do the version bump in Cargo.toml. too hard to do on my phone 😃

oli-obk:
Oh right, we have a changelog. I'm so bad at updating it. Please don't let me stand in your way of adding this to the changelog

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cargo_metadata"
-version = "0.18.0"
+version = "0.19.0"
```
oli-obk, line 3: Ah, since we are pre-1.0, a minor bump is a breaking bump. We can use a patch bump here

oli-obk:
published a new version with this change. Thanks!

PR #255, opened 2023-11-15
Fix docs for `Target.kind` and `Target.crate_types`

Shnatsel:
I only realized that far more values than documented are possible in the `Target.kind` fields after my code broke for end users.

This brings the crate doc commends in line with [the official documentation on `cargo metadata`](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html).

The docs for `kind` are based on both the `cargo metadata` documentation and my experiments, while the docs for `crate_types` are based on documentation alone.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -507,7 +507,7 @@ pub struct Target {
     /// [Cargo crate types](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-crate-type-field):
     /// `bin`, `lib`, `rlib`, `dylib`, `cdylib`, `staticlib`, `proc-macro`
     pub kind: Vec<String>,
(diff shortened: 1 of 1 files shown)


jplatte (commented):
(not a maintainer)

There should probably be a newline in the docs between the first sentence and the rest for both fields (so that it renders as two paragraphs).

oli-obk:
We should probably make these a non_exhaustive enum.

For now, please also document that new variants may get added at any time

Shnatsel:
Yes, I agree that an enum would be better. But that would break semver, and I wanted to start with a less invasive change.

Shnatsel:
I've addressed all review feedback, PTAL.

Issue #239, opened 2023-07-12
use an enum for Target::kind and Target::crate_types


oli-obk:
We know all possible variants, and can still include an `Unknown(String)` fallback to be more forward compatible.

The fields I'm talking about are https://github.com/oli-obk/cargo_metadata/blob/56fcbaaa46e0ff995b4ad16a878db0e1c8c68ba1/src/lib.rs#L440-L446

PR #258, opened 2024-01-20
Converting Target kind and crate_types to enum representation

michaelciraci:
Converting Target kind and crate_types to enum representation, along with an Unknown(String) as a fallback.

This is in response to the following issue: https://github.com/oli-obk/cargo_metadata/issues/239

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -554,44 +554,43 @@ pub struct Target {
 }
 
 impl Target {
-    fn is_kind(&self, name: &str) -> bool {
-        let enum_kind: TargetKind = name.into();
-        self.kind.iter().any(|kind| kind == &enum_kind)
+    fn is_kind(&self, name: TargetKind) -> bool {
+        self.kind.iter().any(|kind| kind == &name)
     }
 
     /// Return true if this target is of kind "lib".
     pub fn is_lib(&self) -> bool {
-        self.is_kind("lib")
+        self.is_kind(TargetKind::Lib)
     }
 
     /// Return true if this target is of kind "bin".
     pub fn is_bin(&self) -> bool {
-        self.is_kind("bin")
+        self.is_kind(TargetKind::Bin)
     }
 
     /// Return true if this target is of kind "example".
     pub fn is_example(&self) -> bool {
-        self.is_kind("example")
+        self.is_kind(TargetKind::Example)
     }
 
     /// Return true if this target is of kind "test".
     pub fn is_test(&self) -> bool {
-        self.is_kind("test")
+        self.is_kind(TargetKind::Test)
     }
 
     /// Return true if this target is of kind "bench".
     pub fn is_bench(&self) -> bool {
-        self.is_kind("bench")
+        self.is_kind(TargetKind::Bench)
     }
 
     /// Return true if this target is of kind "custom-build".
     pub fn is_custom_build(&self) -> bool {
-        self.is_kind("custom-build")
+        self.is_kind(TargetKind::CustomBuild)
     }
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -554,43 +554,158 @@ pub struct Target {
...
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+/// Kind of target.
+///
+/// The possible values are `example`, `test`, `bench`, `custom-build` and
+/// [Cargo crate types](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-crate-type-field):
+/// `bin`, `lib`, `rlib`, `dylib`, `cdylib`, `staticlib`, `proc-macro`.
+///
+/// Other possible values may be added in the future.
```
oli-obk, line 604: Please mark it as non_exhaustive so we can add new variants without it being a breaking change

oli-obk:
Thanks for working on this.

Please update the changelog, too

michaelciraci:
I was looking into why it failed for `nightly`, and it looks like the `main` branch is also failing `nightly`. I have the fix. Do you want me to add it here to get this `nightly` pipeline working again, or open a new merge request?

oli-obk:
Either is fine with me. Whatever works for you

tests/test_samples.rs
```diff
@@ -208,7 +208,7 @@ fn all_the_fields() {
         }
     );
     assert_eq!(meta.workspace_members.len(), 1);
-    assert!(meta.workspace_members[0].to_string().starts_with("all"));
+    assert!(meta.workspace_members[0].to_string().contains("all"));
```
oli-obk, line 211: What changed here? Can we get the original display impl back with a manual impl?

michaelciraci: In `stable`, that string is: `all 0.1.0 (path+file:///Users/michaelciraci/Documents/cargo_metadata/tests/all)`
In `nightly`, that string is: `path+file:///Users/michaelciraci/Documents/cargo_metadata/tests/all#0.1.0`

That's why the nightly tests were failing.

michaelciraci:
It looks like the latest version of [cargo-platform](https://crates.io/crates/cargo-platform) (v0.1.6), has a msrv of 1.70. However v0.1.5 works with 1.56. I put an upper limit on `cargo-platform` to keep the msrv of 1.56.

jplatte:
> I put an upper limit on `cargo-platform` to keep the msrv of 1.56.

That is generally a very bad idea, as it leads to un-resolvable dependency trees when another dependency in the tree requires `cargo-platform = "0.1.6"`. Instead, I would recommend updating the MSRV CI job to run `cargo update -p cargo-platform --precise 0.1.5` before checking compilation. There are also other solutions that don't require touching the CI job (or `Cargo.toml`) whenever a dependency bumps its MSRV in a patch release, but they're more involved; maybe the maintainers have a preference towards one such solution, I'm just a contributor.

michaelciraci:
Yeah, there are multiple ways to deal with this and I'll defer to the maintainers on this.

The pull request was really for something else, so I can revert the `cargo-platform` change and that can be dealt with in a different pull request as well.

oli-obk:
> The pull request was really for something else, so I can revert the `cargo-platform` change and that can be dealt with in a different pull request as well.

yes please remove that commit

oli-obk:
Please remove the commits, instead of using a new commit to revert the previous one

michaelciraci:
Should be good now :)

oli-obk:
Thanks!

PR #260, opened 2024-04-10
use cargo_util_schemas::{PackageName, FeatureName}

aatifsyed:
Testing the waters for how you'd feel about this change

[cargo_metadata](https://docs.rs/cargo_metadata) has validated strings for these usecases.
- this would be a breaking change with a new public dependency

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -304,7 +304,7 @@ pub struct Node {
 pub struct NodeDep {
     /// The name of the dependency's library target.
     /// If the crate was renamed, it is the new name.
-    pub name: String, // TODO(aatifsyed): should this be PackageName?
+    pub name: PackageName,
     /// Package ID (opaque unique identifier)
     pub pkg: PackageId,
     /// The kinds of dependencies.
@@ -358,7 +358,8 @@ pub struct Package {
     pub id: PackageId,
     /// The source of the package, e.g.
     /// crates.io or `None` for local projects.
-    pub source: Option<Source>, // TODO(aatifsyed): should this be RegistryName?
+    // Note that this is NOT the same as cargo_util_schemas::RegistryName
+    pub source: Option<Source>,
     /// The [`description` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-description-field) as specified in the `Cargo.toml`
     pub description: Option<String>,
     /// List of dependencies of this particular package
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,7 @@ pub use camino;
 pub use semver;
 use semver::Version;
 
+use cargo_util_schemas::manifest::{FeatureName, PackageName};
 #[cfg(feature = "builder")]
 pub use dependency::DependencyBuilder;
 pub use dependency::{Dependency, DependencyKind};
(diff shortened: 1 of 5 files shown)


weihanglo (commented):
`cargo_metadata` basically consumes the output JSON from `cargo metadata`, which should be validated by cargo itself already. I am not sure if I see the value of extra validations in this crate. Could you elaborate why this is needed?

oli-obk:
I don't care about the validation, but sharing data types with cargo may be a useful way forward to slowly making cargo_metadata mostly a set of convenience functionality

aatifsyed:
CI is failing because

```toml
[package]
name = "cargo-util-schemas"
version = "0.2.0"
rust-version = "1.75.0"  # MSRV:1
```
https://docs.rs/crate/cargo-util-schemas/0.2.0/source/Cargo.toml.orig

So this would be an MSRV bump as well as an API breaking change

aatifsyed:
Just a gentle bump on this - under consideration or should it be closed?

I will say that having `HashMap<PackageName, semver::Version>` is fantastic UX ootb in my own experiments on this fork :)

oli-obk:
That's quite an MSRV bump. What's the earliest rustc version that supports `rust-version` in `Cargo.toml`? Maybe we can first bump to that and add an MSRV to cargo_metadata. Then when we bump it again in the future, the cargo version resolution code can at least figure out which versions it can use.

weihanglo:
> What's the earliest rustc version that supports `rust-version` in `Cargo.toml`?

[1.56](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#rust-version), though Cargo's MSRV-aware resolver recognizes it even when the version is older than 1.56.
So maybe use [`cargo-msrv`](https://crates.io/crates/cargo-msrv) to find the current MSRV?

oli-obk:
Oh neat, thanks!

oli-obk:
Oh well, let's bump the rust-version. It's a major version bump, and anyone using an old cargo to compile cargo_metadata to use it on a newer crate is already asking for trouble anyway.

src/lib.rs
```diff
@@ -357,7 +358,7 @@ pub struct Package {
     pub id: PackageId,
     /// The source of the package, e.g.
     /// crates.io or `None` for local projects.
-    pub source: Option<Source>,
+    pub source: Option<Source>, // TODO(aatifsyed): should this be RegistryName?
```
aatifsyed, line 361: here

src/lib.rs
```diff
@@ -303,7 +304,7 @@ pub struct Node {
 pub struct NodeDep {
     /// The name of the dependency's library target.
     /// If the crate was renamed, it is the new name.
-    pub name: String,
+    pub name: String, // TODO(aatifsyed): should this be PackageName?
```
aatifsyed, line 307: and here

aatifsyed (commented):
That leaves two TODOs, introducing `PackageName` and `RegistryName`

I'll push a commit adding those changes shall I?

src/lib.rs
oli-obk: Yea, I think we can remove the `Source` type now

src/lib.rs
oli-obk: yea I think PackageName works here

src/lib.rs
aatifsyed: Having seen `Source::is_crates_io`, it's not the same as RegistryName

oli-obk (commented):
Yea I like this. Let's bundle it into the next major release

PR #261, opened 2024-04-28
Enable documentation of non-default features and feature gates

zetanumbers:
I saw some cargo features but haven't find any documentation for those. This PR would become helpful in such situations.

Uses [`doc_auto_cfg`](https://github.com/rust-lang/rust/issues/43781) to generate documentation for feature gates (which is [used by bevy](https://github.com/bevyengine/bevy/pull/12366)). See about how to customize docs.rs build [here](https://docs.rs/about/builds).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ default = []
 builder = ["derive_builder"]
 unstable = []
 
+[package.metadata.docs.rs]
+all-features = true
+
 [package.metadata.cargo_metadata_test]
 some_field = true
 other_field = "foo"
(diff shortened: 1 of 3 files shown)


jplatte (approved):

Cargo.toml
```diff
@@ -23,6 +23,9 @@ default = []
 builder = ["derive_builder"]
 unstable = []
 
+[package.metadata.docs.rs]
+all-features = true
```
jplatte, line 27: Actually isnt `rustdoc-args = ["--cfg", "docsrs"]` also required? Or does docs.rs set the cfg automatically these days?

zetanumbers:
Well, that test failure is not because of me. Somebody raised their minimal rust version for our dependency.

PR #262, opened 2024-05-22
Implement `ops::Index` with `PackageId` for `Resolve` graph

MarijnS95:
While iterating through dependency packages, I find myself needing to look up dependency nodes recursively through the graph, in the same way that I need `Package` information for its `PackageId` via the already-existing `ops::Index` implementation on `Metadata`.  This is simpler if the indexing operation is also available on a `Resolve`d graph.


oli-obk:
cycling CI

oli-obk:
Thanks!

PR #266, opened 2024-10-03
fix(Metadata): add `#[serde(default)]` on workspace_default_members field

zjp-CN:
Set a default value when Metadata is deserialized with WorkspaceDefaultMembers skipped in serialization.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -158,7 +158,7 @@ pub struct Metadata {
     /// The list of default workspace members
     ///
     /// This not available if running with a version of Cargo older than 1.71.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -259,6 +260,14 @@ pub fn workspace_default_members_is_missing(
     workspace_default_members.0.is_none()
 }
 
+/// A default value when Metadata is deserialized with WorkspaceDefaultMembers skipped.
+///
+/// We use this function for now because WorkspaceDefaultMembers doesn't implement Default.
```
oli-obk, line 265: I'd prefer just deriving `Default`

oli-obk:
Thanks

PR #273, opened 2024-11-24
Update dependencies

regexident:


oli-obk:
I was gonna roll it into whatever next actual change comes and release it with that. Or is this actually useful beyond bumping deps and not having duplicates in a dep tree?

smoelius:
> Or is this actually useful beyond bumping deps and not having duplicates in a dep tree?

This is precisely why I want it, actually.

PR #276, opened 2024-12-03
add FromStr and Display impls for TargetKind and CrateType

### LINKED ISSUE #275: Add Display impls for CrateType and TargetKind?
Hey, thanks as always!

I'm wondering if you would be up for adding Display impls for `CrateType` and `TargetKind` to mirror the serde impls.

* It's possible to write this externally but necessarily requires a panic for an unknown variants (due to `#[non_exhaustive]`.
* It's also possible to serialize them into a string via a custom serializer I guess, but a lot of work.

Thanks!

sunshowers:
I'd like to use the string representations in a few places (particularly to avoid exposing cargo-metadata types as a public dependency). Add `FromStr` and `Display` impls.

Closes #275.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,7 +88,7 @@ use std::fmt;
 use std::hash::Hash;
 use std::path::PathBuf;
 use std::process::{Command, Stdio};
-use std::str::from_utf8;
+use std::str::{from_utf8, FromStr};
 
 pub use camino;
 pub use semver;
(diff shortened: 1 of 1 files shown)


sunshowers:
The clippy issue is due to an unrelated new lint in Rust 1.83, and the 1.73 test issue is due to one of the dependencies bumping its MSRV.

oli-obk:
Slightly sad about the duplication of string <-> variant mapping, but not worth a macro yet.

PR #277, opened 2024-12-03
Make `Dependency`'s `source` member the same type as `Package`'s `source` member: `Option<Source>`

regexident:
They have the exact same representation (i.e. bare string) in the JSON output of `cargo metadata`, so might as well use the same type in `Metadata`.

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -8,6 +8,8 @@ use derive_builder::Builder;
 use semver::VersionReq;
 use serde::{Deserialize, Deserializer, Serialize};
 
+use crate::Source;
+
 #[derive(Eq, PartialEq, Clone, Debug, Copy, Hash, Serialize, Deserialize, Default)]
 /// Dependencies can come in three kinds
 pub enum DependencyKind {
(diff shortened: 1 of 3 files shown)


src/dependency.rs
```diff
@@ -83,3 +83,5 @@ pub struct Dependency {
 }
 
 pub use cargo_platform::Platform;
+
+use crate::Source;
```
oli-obk, line 87: 😆 I have only myself to blame for the import location

oli-obk:
This will need a breaking change release 🤔

src/dependency.rs
regexident: Fixed.

regexident:
> This will need a breaking change release 🤔

It will. It's of no urgency though, so feel free to postpone merging at your discretion.
I just stumbled upon the inconsistency and figured you might appreciate a PR, nothing that's actively in the way.

regexident:
@oli-obk With the already merged [update from `cargo_platform` `0.1.9` to `0.2.0`](https://github.com/oli-obk/cargo_metadata/pull/289), which (due to exposing its `Platform` type and a MSRV bump) will already be a semver breaking change, would you mind merging this as well now?

PR #271, opened 2024-11-20
Derive `Hash` traits for `Metadata` and its transitive members

regexident:
This PR is sort of a follow-up of https://github.com/oli-obk/cargo_metadata/pull/267, aiming to fill the very last remaining blind spot: `Hash`.

(Motivation: I found myself in need of `Hash` for `Metadata` and its members as I otherwise couldn't use them as `HashMap`/`HashSet` keys, which is crucial for my use case.)

PR #280, opened 2024-12-04
build: fix -Zminimal-versions

paolobarbolini:
`serde_json::Value` started implementing `Hash` starting from v1.0.118

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ cargo-platform = "0.1.2"
 derive_builder = { version = "0.20", optional = true }
 semver = { version = "1.0.7", features = ["serde"] }
 serde = { version = "1.0.136", features = ["derive"] }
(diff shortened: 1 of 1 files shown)


oli-obk:
While you're at it, can you add this flag to the nightly CI?

weihanglo:
You might want [direct-minimal-versions](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#direct-minimal-versions) instead

ia0:
Thanks for the PR! I just hit this issue. For reference this was introduced by #271. And we could argue that `serde_json` should have bumped the minor version for that, but that's orthogonal.

paolobarbolini:
> And we could argue that `serde_json` should have bumped the minor version for that, but that's orthogonal.

I don't know why but David never bumps minor version.

Issue #279, opened 2024-12-03
Provide higher-level API around a `Package`'s features?


regexident:
Before opening a PR for this I wanted to first check if such a change would be aligned with your vision of this crate, @oli-obk.

### Motivation

With the transitive required features of a package encoding semantic information (weak vs. strong, dep vs. feature, transitive vs. non-transitive etc.) these days it feels like `cargo_metadata` should provide a fail-safe and higher-level API for that, rather than expecting users to manually perform pattern matching on the bare string representation, potentially fragmenting the ecosystem in the long-term due to subtle bugs downstream.

### Draft implementation

Change the following member of `Package`:

```rust
pub struct Package {
    // ...

    // before:
    pub features: BTreeMap<String, Vec<String>>,
    // after:
    pub features: BTreeMap<String, Vec<FeatureDep>>,

    // ...
}
```

Introduce a new-type around the map's value type, that could look something like this:

```rust
/// Dependency of a crate's feature.
#[derive(Clone, Serialize, Deserialize, Debug, PartialOrd, Ord, PartialEq, Eq, Hash)]
#[serde(transparent)]
pub struct FeatureDep {
    /// The underlying string representation.
    pub repr: String,
}

impl FeatureDep {
    /// Returns true if it is a dependency feature
    /// (i.e. `"dep:<dependency>"`).
    pub fn is_dependency(&self) -> bool {
        self.repr.starts_with("dep:")
    }

    /// Returns true if it is a transitive feature of a dependency
    /// (i.e. `"<dependency>/<feature>"` or `"<dependency>?/<feature>"`).
    pub fn is_transitive(&self) -> bool {
        self.repr.contains("/")
    }

    /// Returns true if it is a weak transitive feature of an optional dependency
    /// (i.e. `"<dependency>?/<feature>"`).
    pub fn is_weak(&self) -> bool {
        self.repr.contains("?/")
    }

    /// Returns the name of the package dependency if it is a dependency feature
    /// (i.e. `"dep:<package>"`, `"<package>/<feature>"` or `"<package>?/<feature>"`).
    ///
    /// Note: For dependency features without explicit `"dep:"` prefix this will return `None`.
    pub fn package(&self) -> Option<String> {
        if let Some(package) = self.repr.strip_prefix("dep:") {
            return Some(package.to_owned());
        }

        if let Some((package, _)) = self.repr.split_once("/") {
            if let Some(package) = package.strip_suffix("?") {
                return Some(package.to_owned());
[...]

PR #281, opened 2024-12-04
A visitor API for exploring a package's features and their dependencies

regexident:
An exploration for how `Metadata` could provide a flexible API for processing a package's feature dependency graph.

Resolves https://github.com/oli-obk/cargo_metadata/issues/279


src/visit.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    /// Visits a missing dependency.
+    ///
+    /// Return `Ok(())` to continue the walk, or `Err(…)` to abort it.
+    fn visit_missing_dependency(&mut self, dep_name: &str) -> Result<(), Self::Error>;
+
+    /// Visits a missing package.
+    ///
+    /// Return `Ok(())` to continue the walk, or `Err(…)` to abort it.
+    fn visit_missing_package(&mut self, pkg_name: &str) -> Result<(), Self::Error>;
```
obi1kenobi, line 18: I'm confused about the terminology of `missing dependency` and `missing package`. In what sense are they missing, given that we seem to know their names based on the function signatures?

While all the trait items are documented, it would be good to frame their documentation in terms of "what does the user need to understand to choose a good implementation here." Right now the documentation seems to be written from the perspective of "what does the code *driving* the visitor need to know to drive it properly", which isn't that helpful for end-users.

src/visit.rs
```diff
@@ -0,0 +1,211 @@
...
+    ///
+    /// Return `Ok(<walk>)` where `<walk>` indicates whether or not to walk the feature's downstream dependencies,
+    /// or `Err(…)` to abort the walk.
+    fn visit_dep_feature(
+        &mut self,
+        package: &Package,
+        dep_name: &str,
+        dep_feature: &str,
+        weak: bool,
+    ) -> Result<bool, Self::Error> {
```
obi1kenobi, line 52: The example syntax is `crate_name/feat_name` but neither of those names appears in the list of function arguments. Could you make the example match the function arguments?

src/visit.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Visits a feature that's enabling a dependency with `dep:dep_name` syntax.
+    ///
+    /// Return `Ok(<walk>)` where `<walk>` indicates whether or not to walk the feature's downstream dependencies,
+    /// or `Err(…)` to abort the walk.
+    fn visit_dep(&mut self, package: &Package, dep_name: &str) -> Result<bool, Self::Error> {
+        let (..) = (package, dep_name);
+        Ok(true)
+    }
+
+    /// Visits a feature that's enabling a feature on a dependency with `crate_name/feat_name` syntax.
```
obi1kenobi, line 42: Both of these doc comments say "visits a feature ..." but they both seem to take `package: &Package` and nothing resembling a feature. I'm confused about how this interface works =/

src/dependency.rs
```diff
@@ -82,4 +82,11 @@ pub struct Dependency {
     pub path: Option<Utf8PathBuf>,
 }
 
+impl Dependency {
+    /// The dependency's possibly renamed name.
+    pub fn renamed(&self) -> &str {
+        self.rename.as_deref().unwrap_or(&self.name)
+    }
+}
```
obi1kenobi, line 90: `renamed` doesn't seem like the right name for this concept. It's strange that we can use a `renamed()` accessor on dependency that isn't renamed, and get anything other than `None`.

obi1kenobi (commented):
Personally, I'm unfortunately not a fan of this API. I find it very confusingly structured and documented, and it isn't clear to me that the flexibility a visitor offers compared to other approaches is valuable for many use cases.

I can imagine two modes where feature information is useful:
- Which things are directly enabled by which features? This is analogous to "what's written in the Cargo.toml next to each feature."
- What are all the things enabled by each feature? This is analogous to "if I enable feature X, what do I actually get."

Obviously, a visitor can cover both modes and everything in between. But it isn't clear to me that there is _any use case_ that lies in between. So I can't justify the complexity of either maintaining this API, nor of users learning how to use it.

For my 2 cents, I'd prefer an API that offered functionality for each of the two modes, and was designed to be simple and easy to understand.

src/visit.rs
regexident: The documentation was mostly copied from the corresponding APIs on `cargo` itself, but the names used in the implementation slightly drifted afterwards. I've cleaned things up and also added more information. 👍 

The need for `visit_missing_dependency(…)` and `visit_missing_package(…)` is twofold:

1. The `Metadata` only contains `packages` for crates that were part of the active crate graph, i.e. crates that either were workspace members, non-optional dependencies or optional dependencies that where enabled via corresponding features. So if you walk the feature graph from a feature that would enable a dependency but that dependency wasn't also activated via `MetadataCommand.features(…)`, then lookup of the corresponding `Package` on `Metadata` will fail. But we need the `Package` to look up its features for the recursion. (I'd expect `visit_missing_dependency(…)` to be unnecessary for valid `Metadata` instances and haven't seen it fail myself so far.)

2. The `Metadata` might have been deserialized from an unchecked JSON blob with packages/dependencies missing. I'll leave deciding on whether or not we want to gracefully handle such situations to @oli-obk, but wanted to avoid pestering the draft implementation with `.unwrap()`s.

These failure modes are also partly what made me switch to a visitor API. 

I had initially implemented a single method on `Metadata` but the implementation very quickly became unwieldy and inflexible:

```rust
fn transitive_feature_dependencies<'a>(&'a self, package: &'a Package, feature_name: &str) -> Result<BTreeMap<&'a Package, BTreeSet<FeatureValue>>> {
    // ...
}
```

But whether or not this method should fail on missing packages depends on the use-case, I think.
With a visitor this decision is totally up to you, without bloating the crate's API with options.

Similarly one may want to be able to pick between different collection modes: recursive (i.e. transitive) and non-recursive (i.e. immediate) feature dependencies. Again, a visitor API leaves this up to the user, without bloating the crate's API with options.

src/visit.rs
regexident: Fixed.

src/visit.rs
regexident: The three visitor methods correspond to the variants of the `FeatureValue` type (which was copied from `cargo` itself). Every time the walker passes such a value in the feature graph it calls the corresponding method on the visitor that it carries along.

The variants of said type unfortunately have rather confusing names already: `Feature`, `Dep`, `DepFeature`, especially when considering that `FeatureValue` is itself the value of a "feature dependency" (i.e. a feature depended on by another feature). Finding proper names for these was … a challenge. One that I decided to mostly gloss over for now. #bike-shedding.

Preferably the `FeatureValue` enum would have new-type variants, rather than struct variants, so that each variant's type could be passed to the corresponding visitor method and also so that the corresponding type of feature value could be handled as a single value. But! By promoting each variant to individual types each of them would have to have their own `Serialize`/`Deserialize`/`FromStr` implementations. An explosion of complexity and change that I wanted to avoid if possible. That said I do think that the end goal should be a new-type based enum.

obi1kenobi:
Pulling your comment [out of the thread](https://github.com/oli-obk/cargo_metadata/pull/281#discussion_r1871679081) for broader visibility:

> But whether or not this method should fail on missing packages depends on the use-case, I think.
With a visitor this decision is totally up to you, without bloating the crate's API with options.
>
> Similarly one may want to be able to pick between different collection modes: recursive (i.e. transitive) and non-recursive (i.e. immediate) feature dependencies. Again, a visitor API leaves this up to the user, without bloating the crate's API with options.

I don't find this argument convincing, personally. Leaving more things up to the user is not always a good choice.

`cargo_metadata` already has low-level functionality that supports all the possible special cases and leaves maximum choice up to the users. I don't think it's worth adding a new and very complex layer that also tries to be super generic.

I think what's missing is a _simple_ way to answer the most common questions:
- What is directly enabled by this feature?
- What are all things transitively enabled by this feature?

Something a user could easily find, quickly understand, and immediately use in their own code. A visitor API in my book accomplishes none of that.

I'm not the `cargo_metadata` maintainer so the final call isn't mine to make. But as a user of `cargo_metadata`, I wouldn't switch from what I currently have to a visitor-based API. It would make everything _more complex_ for me to understand and maintain, not less.

regexident:
> I'm not the `cargo_metadata` maintainer so the final call isn't mine to make. But as a user of `cargo_metadata`, I wouldn't switch from what I currently have to a visitor-based API. It would make everything _more complex_ for me to understand and maintain, not less.

And I wouldn't expect you to. For the 80% I'd expect convenient methods to exist. The for other 20% a visitor API is preferable.

One such use case being mine: I'd need to be able to selectively block certain downstream features from (it and its dependencies) being collected. With a method that only returns the immediate dependencies I'd be out of luck as I would have to write the recursive traversal logic myself, which is arguably more difficult (to do right) than implementing a visitor.

oli-obk:
I also feel like the visitor is overkill, but until we get generators, I can also see how an iterator based version may be annoying to implement.

I'd be fine adding this under an `unstable_visitor` feature gate, and using it to implement a higher level API (along with giving you the ability to play with it directly), but I'm hesitant to support a visitor API forever.

PR #282, opened 2024-12-10
Add method to check whether WorkspaceDefaultMembers has a value

### LINKED ISSUE #254: Don't panic from WorkspaceDefaultMembers?
Hi, https://docs.rs/cargo_metadata/0.18.1/cargo_metadata/struct.WorkspaceDefaultMembers.html documents that it will panic on cargo <1.70, and it does.

For cargo-mutants, I'd like to use this but also support older versions of cargo, so it would be really nice if this was a runtime error rather than a panic. Or at least, if there was a documented version field that I could check before looking at it. Possibly I am missing some subtlety about why that's not feasible.

Thanks for a really helpful crate.

foresterre:
This method checks whether the called `cargo-metadata` supports reading the workspace default members, which is false for versions of Cargo prior to 1.71

Alternatively, I could:
- add an Option variant method like `workspace_default_packages` directly on `Metadata`: `pub fn workspace_default_packages(&self) -> Option<Vec<&Package>>`, or
- reverse the name from `is_available` to `is_missing` (which would also allow the removal of the `doc(hidden)` function `workspace_default_members_is_missing`).

fixes #254

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,8 +157,11 @@ pub struct Metadata {
     pub workspace_members: Vec<PackageId>,
     /// The list of default workspace members
     ///
-    /// This not available if running with a version of Cargo older than 1.71.
-    #[serde(default, skip_serializing_if = "workspace_default_members_is_missing")]
+    /// This is not available if running with a version of Cargo older than 1.71.
+    ///
+    /// You can check whether it is available or missing using respectively
+    /// [`WorkspaceDefaultMembers::is_available`] and [`WorkspaceDefaultMembers::is_missing`].
+    #[serde(default, skip_serializing_if = "WorkspaceDefaultMembers::is_missing")]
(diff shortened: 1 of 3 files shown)


oli-obk:
> * reverse the name from `is_available` to `is_missing` (which would also allow the removal of the `doc(hidden)` function `workspace_default_members_is_missing`).

is this test usually going to be used as `if !foo.is_available() {...}` or as `if foo.is_available() {}`? Oh to be honest we could just add both, just like `Option` has `is_none` and `is_some`.

foresterre:
I've updated the PR to include both. 

I've also removed the `doc(hidden) workspace_default_members_is_missing`, which is no longer necessary when the is_missing variant exists. It was both marked hidden and noted that it "might be removed in the future". Are you okay with its removal?

PR #284, opened 2025-02-23
feat: `2024` edition stabilized

HTGAzureX1212:
This PR renames the `Edition::_E2024` variant to `Edition::E2024` and exposes it in the documentation given its stabilization in Rust 1.85.0.


oli-obk:
Can you also open a PR bumping the patch version? Then I can publish a new version from my phone

fenhl:
#285

PR #285, opened 2025-03-01
Version bump

fenhl:
Bumps the patch version as requested in https://github.com/oli-obk/cargo_metadata/pull/284#issuecomment-2688912084.


oli-obk:
Thanks! new version released

PR #289, opened 2025-04-26
Update cargo-platform to 0.2.0

UebelAndre:


UebelAndre:
It would also be great if a release could be cut after this change!

oli-obk:
needs an MSRV bump and please also increase the patch version number

UebelAndre:
> needs an MSRV bump and please also increase the patch version number

Done!

oli-obk:
oops sorry forgot about this

Issue #294, opened 2025-05-29
Old version of `cargo-util-schemas` used


smoelius:
Commit e3373d02e79dc64adbecb3fe32fecc1dd324bba6 introduced a somewhat old version of `cargo-util-schemas`:
https://github.com/oli-obk/cargo_metadata/blob/fa60b98bc8016ee4702b2512dbb1b0d68669cdb0/Cargo.toml#L15

The most recent published version is [0.8.1](https://crates.io/crates/cargo-util-schemas/0.8.1).

Would you accept a PR to update the version that `cargo_metadata` uses?

smoelius:
> Jup, tho with recent feedback we may just switch to using newtype wrappers defined in cargo_metadata itself

Sorry, what are you referring to?

TBH, I wanted `cargo_metadata` to update to the latest version of `cargo-util-schemas` so that I could push a change to their newtype wrappers and have it reflected in `cargo_metadata`.

The specific change I wanted to incorporate was a more general `PartialEq` implementation, so, e.g., `package.name == "my-package"` would work:
```rs
        impl<T: AsRef<str>, Rhs: AsRef<str>> PartialEq<Rhs> for $name<T> {
            fn eq(&self, other: &Rhs) -> bool {
                self.as_ref() == other.as_ref()
            }
        }
```

oli-obk:
> Sorry, what are you referring to?

https://github.com/oli-obk/cargo_metadata/issues/63#issuecomment-2911517148

tamird:
+1 to avoiding (or trimming) cargo-util-schemas, which pulls in quite a lot of stuff.

smoelius:
Is "verification" one of the benefits the newtype wrappers provide to `cargo_metadata`? https://github.com/rust-lang/cargo/blob/e647065d5de87f614823df6b29db2b3adf83b085/crates/cargo-util-schemas/src/manifest/mod.rs#L1367

I'm asking because that logic is in [restricted_names.rs](https://github.com/rust-lang/cargo/blob/e647065d5de87f614823df6b29db2b3adf83b085/crates/cargo-util-schemas/src/restricted_names.rs), and I'm wondering if it will need to be incorporated into `cargo_metadata`.

PR #299, opened 2025-06-18
Re-export PackageName from cargo-util-schemas

kornelski:
This crate already exposes `PackageName` from cargo-util-schemas, but the type can't be named without also importing the cargo-util-schemas crate.

This PR re-exports the `PackageName`, so that users don't have to include cargo-util-schemas themselves, and hopefully it will be easier to upgrade it in the future.

BTW: currently cargo-util-schemas version used is 0.2, but the latest is 0.8, and 0.7 is the latest MSRV-compatible one.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,7 +100,8 @@ pub use camino;
 pub use semver;
 use semver::Version;
 
-use cargo_util_schemas::manifest::{FeatureName, PackageName};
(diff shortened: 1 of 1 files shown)


oli-obk:
As discussed in https://github.com/oli-obk/cargo_metadata/issues/294 depending on cargo-util-schemas may not be worth it (there are no useful methods on the types), so if we're changing anything here we could just as well use our own newtypes

PR #300, opened 2025-06-28
fixes "package name cannot be empty" under "resolve/nodes/deps/name"

joeferner:
When using "-Z bindeps" with locally referenced "build-dependencies" the "name" field is an empty string. PackageName reqires a non-empty string. Cargo's cargo_output_metadata.rs Dep struct's "name" field is defined as "InternedString" which does not have this restriction.

Fixes https://github.com/rust-lang/rust-analyzer/issues/20002

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,7 +344,7 @@ pub struct Node {
 pub struct NodeDep {
     /// The name of the dependency's library target.
     /// If the crate was renamed, it is the new name.
-    pub name: PackageName,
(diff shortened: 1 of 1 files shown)


joeferner:
Added some comments, hope this clarifies things.

zzjrabbit:
It seems that although this pull request can be cleanly merged, it hasn't been yet. But this pull request is important, since it can help rust-analyzer fix bugs in bindeps.

Issue #63, opened 2019-01-16
How can Cargo help


Eh2406:
At the Cargo Team Meeting today we discussed the importance of crates like this one. We decided to reach out to discuss how best to work together. How can we insure that the projects are kept smoothly uptodate with each other? What else can the Cargo team do to make sure we don't accidentally break this project? How can the Cargo team make sure that changes get included here promptly? Can we / Should we share code?

cc @ehuss who already got started in https://github.com/oli-obk/cargo_metadata/pull/61
and @dwijnand had some suggestions in https://github.com/rust-lang/cargo/pull/6534

Also this idea should be expanded to Cargos other outputs. Should that be done here, or in some other existing crate, or in some new Cargo Team maintained crate?

cc @epage as this relates to the discussion in https://github.com/oli-obk/cargo_metadata/issues/45

epage:
> Also this idea should be expanded to Cargos other outputs. Should that be done here, or in some other existing crate, or in some new Cargo Team maintained crate?

#53 and https://github.com/crate-ci/escargot/ have `cargo build` output.

On a related note, `--message-format` for `cargo test` doesn't have any value from what I can see because it will print the messages for the build but then switch to dumping the test programs output without any delimiter.  I'd love to have a way to get actual test failure information programmatically.

ehuss:
> test failure information programmatically

Are you talking about the libtest json output or something different?  i.e., today you can get test information with `cargo +nightly test --message-format=json -- --format=json -Zunstable-options`.  It's not stable, and it's not clear what it's future will look like with custom test frameworks.  I also wish it included more information like filename/line numbers.

epage:
Was not aware of that unstable feature, thanks!  I can understand the challenge of stablizing it.

oli-obk:
I think sharing the datastructures in a common crate would solve most compatibility issues. If cargo does a backwards compatible change to the datastructures, everyone gets the update for free (via minor bumps), if it's a breaking change, we also bump `cargo_metadata`.

For the diagnostics json datastructures I also want to push for rustc to share the crates.io datastructures.

dwijnand:
A few questions:

1. Should the crates share the same versioning as cargo or not?

2. ... and if they're the same version keeping them in the cargo repo makes more sense (to me), but they don't, should they be in the cargo repo or not? 

3. ... and if they're in the same repo, do they sync releases or can they be released independently?

[...]

PR #304, opened 2025-08-19
Add newtype wrapper

smoelius:
Adds a `str_newtype` macro (similar to [cargo-util-schema's](https://github.com/epage/cargo/blob/d8975d2901e132c02b3f6b1d107f2f50b275a058/crates/cargo-util-schemas/src/manifest/mod.rs#L1355-L1413)) and eliminates cargo-util-schema as a dependency.

These changes were discussed in https://github.com/oli-obk/cargo_metadata/issues/63#issuecomment-2911517148 and https://github.com/oli-obk/cargo_metadata/issues/294#issuecomment-3182993049.

If it is helpful, below are two screenshots of diffs between cargo-util-schema's `str_newtype` at commit `d8975d2` and the one introduced in this PR.

Screenshot From 2025-08-18 20-47-34
Screenshot From 2025-08-18 20-47-10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,7 +101,6 @@ pub use cargo_platform;
 pub use semver;
 use semver::Version;
 
-use cargo_util_schemas::manifest::{FeatureName, PackageName};
 #[cfg(feature = "builder")]
 pub use dependency::DependencyBuilder;
 pub use dependency::{Dependency, DependencyKind};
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -129,6 +128,92 @@ mod errors;
...
+macro_rules! str_newtype {
+    ($name:ident) => {
+        /// String newtype
+        ///
+        /// Based on [cargo-util-schema's string newtype] but with two crucial differences:
+        ///
+        /// - This newtype does not verify the wrapped string.
+        /// - This newtype allows comparison with arbitrary types that implement `AsRef<str>`.
+        ///
+        /// [cargo-util-schema's string newtype]: https://github.com/epage/cargo/blob/d8975d2901e132c02b3f6b1d107f2f50b275a058/crates/cargo-util-schemas/src/manifest/mod.rs#L1355-L1413
```
oli-obk, line 140: This documentation is too general imo. I think we should add the docs to the macro invocation and tune it for the specific types so we can link to Cargo.toml documentation

oli-obk: Yea, that's good. Tho i was hoping https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html to have some docs on these, I couldn't find any

smoelius:
@oli-obk Are you waiting for me to rebase? (Sorry for my confusion.)

src/lib.rs
oli-obk: Ah fun. It'll make its way to stable then. I guess just mark the test as ignore with a comment linking to this PR

smoelius:
> Uh no, no idea what I was waiting for Let's disable that test and then I'll merge and release a new version

I should have been more clear. The test was fixed by https://github.com/oli-obk/cargo_metadata/pull/305/commits/d64b6e00d53320bd2aeba1393b4832842346d1cf in #305. So I think all that was needed is a rebase.

PR #305, opened 2025-09-07
Proposed `-Zbuild-dir` fix

smoelius:
This is my proposed fix for the recent nightly breakage (https://github.com/rust-lang/cargo/commit/2b71fd6424de387502b37556c9fb00f657c70c78).

If I understand correctly, `-Zbuild-dir` is not yet in the stable version of Cargo, but it has been stabilized in the nightly version of Cargo.

CC @ranger-ross, who I think introduced the `build_dir` and `build_dir_disabled` tests.

### DIFF
--- a/tests/selftest.rs
+++ b/tests/selftest.rs
@@ -84,11 +84,13 @@ fn builder_interface() {
 
 #[test]
 fn error1() {
+    let manifest_path = current_dir().unwrap().join("foo");
+    let error = "error: the manifest-path must be a path to a Cargo.toml file";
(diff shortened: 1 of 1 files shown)


smoelius:
The other failure appears to be due to this: https://github.com/rust-lang/cargo/commit/db6891f2446260c62c62b9c5a43d0a01f017324f

Fixing it on nightly causes stable to break, though.

I suspect you've run into problems like this before. I'll pause so you can tell me your strategy for handling them.

PR #318, opened 2026-02-28
refactor: allow for more flexible options

dybucc:
The prior signature for the `other_options()` method on `MetadataCommand` made for some slightly awkward conversions that required possibly using an iterator, followed by a `map()` at call site if say, you had `&str`s instead of `String`s. This PR tries to move that out of user code into the library, so as to allow the following,

```rust
let workspace_metadata = MetadataCommand::new()
    .no_deps()
    .other_options(["--format-version=1"]);
```

instead of (possibly, it's just one way)

```rust
let workspace_metadata = MetadataCommand::new()
    .no_deps()
    .other_options([String::from("--format-version=1")]);
```

or

```rust
let workspace_metadata = MetadataCommand::new()
    .no_deps()
    .other_options(["--format-version=1"].into_iter().map(ToString::to_string).collect());
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1118,8 +1118,11 @@ impl MetadataCommand {
     }
     /// Arbitrary command line flags to pass to `cargo`.  These will be added
     /// to the end of the command line invocation.
-    pub fn other_options(&mut self, options: impl Into<Vec<String>>) -> &mut MetadataCommand {
-        self.other_options = options.into();
+    pub fn other_options(
+        &mut self,
+        options: impl IntoIterator<Item: Into<String>>,
+    ) -> &mut MetadataCommand {
+        self.other_options = options.into_iter().map(Into::into).collect();
         self
     }
 


src/lib.rs
```diff
@@ -1118,8 +1118,11 @@ impl MetadataCommand {
     }
     /// Arbitrary command line flags to pass to `cargo`.  These will be added
     /// to the end of the command line invocation.
-    pub fn other_options(&mut self, options: impl Into<Vec<String>>) -> &mut MetadataCommand {
-        self.other_options = options.into();
+    pub fn other_options(
+        &mut self,
+        options: impl IntoIterator<Item: Into<String>>,
```
oli-obk, line 1123: What are your thoughts on replicating what Command::arg does? https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.args

So `AsRef<OsStr>`?

dybucc: The thing is, that could make `other_options()` return a result or otherwise have to panic (probably not a great idea) to convert back into the expected `String` type in the type of the field we're assigning to.

This crate already uses `camino` for UTF-8-compliant paths, so maybe it's alright if we leave it this way.

We could go for a lossy conversion with `to_string_lossy()` on the `OsStr`, but that could replace invalid UTF-8 byte sequences with the replacement character (as per the docs.) And either way, the one thing (I believe) the method tries to communicate is that we need something that can be made into a `String`, and does not lose any information in the conversion.

I'm personally not sure how many folks would query `cargo metadata` with arbitrary input (maybe from a user's code through a command-line option?)

dybucc: The only other option that doesn't have a specific method for in `cargo_metadata` *and* picks up a path is `--lockfile-path`, though it's only in nightly.

Maybe somebody is using the crate for that one option. Either way, I think it's alright to expect the library user to validate their own strings, as we only act as a middle-man to ease access to the information returned by the actual cargo subcommand.

oli-obk: Yea makes sense. We could change the field to OsString to not have any such checks, but everything else is already utf8 as you mention

oli-obk, line 1123: Looks like you need to add a generic parameter instead of using `Item: Into<String>` or we need to bump MSRV

dybucc: Done. Still, is that really it? Maybe I'm missing something, but the CI seems to be failing because of a dependency resolution error.

dybucc: @oli-obk Just committed a PR that should solve the issue with `cargo-platform` not having a compatible MSRV.

oli-obk: Pinning a dep like that will make ppl using both cargo_metadata and cargo-platform very unhappy. 

Let's just bump msrv

dybucc: Done.

dybucc: Tests fail now on `beta` and `nightly` because of a recent change in `cargo`,
where the wording for error messages got changed [[1]]. `stable` seems yet
unaffected so the current MSRV toolchain is fine, but that does mean we would
have to either wait for this to stabilize with a failing CI or otherwise make
the tests more resilient to changes upstream.

[1]: https://github.com/rust-lang/cargo/commit/0e8570b1ec6858229e014a74c3db63ed9fbd1dfc#diff-42736178fb0cc89f7dec03f92640edb16238d40d83ff73b3ec95fa748c4a74e5L1086-L1093

oli-obk:
Hmm if you can find a wording that makes it work on all versions that's obviously ideal, but otherwise just using the nightly/beta version and disabling the test for now (with an issue for reenabling it once our msrv is high enough) works for me

dybucc:
I went for a quick fix for that specific test, where I included among the
strings to match the new one in beta/nightly. Maybe a better idea would be to
scan for the process' error codes, but `cargo` returns 101 on failure no matter
what, so maybe that's not a great idea. There's also the fact that for
integration tests there is no (non-trivial) way of fetching the stdout of the
`cargo` process launched by `cargo_metadata`. I opened a separate PR for the
bump changes.

dybucc:
@oli-obk Not meaning to intrude, but I was wondering if there's anything left to
do?

oli-obk:
Nah, I just forgot. Thanks for the ping

PR #320, opened 2026-04-16
Fill gaps in builders, making them fully usable

regexident:
The crate has an unfortunate limitation with its builder approach since the introduction of `PackageName` and `FeatureName` new-type wrappers: The `builder` feature flag exists and compiles, but a few new-type wrappers block practical use (i.e. `Metadata` cannot currently be created via builders):

1. `PackageName`: `PackageBuilder::name()` requires `Into<PackageName>`, but `PackageName` has no public constructor and doesn't implement From<&str> or From<String>. You'd need `impl From<String> for PackageName` (or at minimum `PackageName::new(s: impl Into<String>)`).

2. `FeatureName`: `NodeBuilder::features()` requires `Vec<FeatureName>`, but `FeatureName` has the same problem: no `From<String>` impl, no public constructor. You'd need `impl From<String> for FeatureName`.

3. `WorkspaceDefaultMembers`: `MetadataBuilder::workspace_default_members()` requires `Into<WorkspaceDefaultMembers>`, but this type wraps `Option<Vec<PackageId>>` with a private field, no `From<Vec<PackageId>>` impl, and no public constructor. You'd need `impl From<Vec<PackageId>> for WorkspaceDefaultMembers`.

4. `PackageId`: This is an opaque string wrapper with no public constructor. You need either `PackageId::new(s: impl Into<String>)` or `impl From<String> for PackageId`.

In short: the `derive(Builder)` generates setters that accept `Into<FieldType>`, but the newtypes used for field types lack `From<…>` impls for their natural input types.

This PR adds `From<String>`/`From<&str>` to `PackageName`, `FeatureName`, `PackageId`, and `From<Vec<PackageId>>` to `WorkspaceDefaultMembers`, making the builder pattern fully usable (e.g. for test construction).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ### Changed
 
-- n/a
+- Added `From<String>` and `From<&str>` for `FeatureName`, `PackageName`, and `PackageId`.
+- Added `From<Vec<PackageId>>` for `WorkspaceDefaultMembers`.
 
 ### Removed
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,6 +173,18 @@ macro_rules! str_newtype {
             }
         }
 
+        impl From<String> for $name<String> {
+            fn from(value: String) -> Self {
+                Self::new(value)
+            }
+        }
+
+        impl<'a> From<&'a str> for $name<String> {
+            fn from(value: &'a str) -> Self {
+                Self::new(value.to_owned())
+            }
+        }
+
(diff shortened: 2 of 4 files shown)


regexident:
@oli-obk I noticed that there's another builder-related PR sort of in limbo: https://github.com/oli-obk/cargo_metadata/pull/316

I'll happily open a release-prep PR, once these (or any other queued) PRs are merged. Just tell me what the version should be: `0.23.2` (since it fixes a breakage from `0.23.0`), or `0.24.0` (since it adds functionality)?

